PR #1, opened 2016-07-01
fixes to work with clang

hannesm:


hannesm:
that's not safe for gcc, sorry (it needs `-Wno-maybe-uninitialized`), do not merge yet

hannesm:
use #2 instead for now

PR #4, opened 2016-07-13
no need for customised version of openlibm

hannesm:
AFAICS the patches from @talex5 were merged upstream, and we can simply use the latest upstream version.  this compiles fine for me on FreeBSD.... (but certainly, additional testing would be great to have)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ Makeconf:
 TOP=$(abspath .)
 FREESTANDING_CFLAGS+=-isystem $(TOP)/nolibc/include
 
-OPENLIBM_URL=https://github.com/talex5/openlibm/archive/v0.4.1-tal1.tar.gz
+OPENLIBM_URL=https://github.com/JuliaLang/openlibm/archive/v0.5.1.tar.gz
 openlibm.tar.gz:
 	curl -sL -o $@ $(OPENLIBM_URL)
 
@@ -40,7 +40,7 @@ build/ocaml/byterun/caml/version.h: build/ocaml/config/Makefile
 	cp config/version.h $@
 
 OCAML_CFLAGS=-O2 -Wall -USYS_linux -DHAS_UNISTD $(FREESTANDING_CFLAGS)
-OCAML_CFLAGS+=-I$(TOP)/build/openlibm/src
+OCAML_CFLAGS+=-I$(TOP)/build/openlibm/include -I$(TOP)/build/openlibm/src
 build/ocaml/asmrun/libasmrun.a: build/ocaml/config/Makefile build/openlibm/Makefile $(OCAML_EXTRA_DEPS)
 	$(MAKE) -C build/ocaml/asmrun \
 	    UNIX_OR_WIN32=unix \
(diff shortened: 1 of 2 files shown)


talex5:
Yes, I think all my patches have been in upstream for a while now.

hannesm:
certainly a similar change should be done in our xen backend whereever openlibm is fetched (I forgot in which repo/opam package that is happening -- xen-minios minios mirage-xen)?

talex5:
Maybe we can switch the Xen backend to use ocaml-freestanding too?

mato:
@hannesm This fails for me with:
````
gcc -I../byterun -DCAML_NAME_SPACE -DNATIVE_CODE -DTARGET_amd64 -DMODEL_default -DSYS_linux  -O2 -Wall -USYS_linux -DHAS_UNISTD -nostdinc -isystem /usr/lib/gcc/x86_64-linux-gnu/5/include -ffreestanding -mno-red-zone -mno-3dnow -I/home/mato/.opam/mirage-solo5-dev/lib/pkgconfig/../../include/solo5-kernel-ukvm/include -isystem /home/mato/solo5/ocaml-freestanding/nolibc/include -I/home/mato/solo5/ocaml-freestanding/build/openlibm/src   -c -o major_gc.o major_gc.c
In file included from major_gc.c:17:0:
/home/mato/solo5/ocaml-freestanding/nolibc/include/math.h:1:22: fatal error: openlibm.h: No such file or directory
compilation terminated.
<builtin>: recipe for target 'major_gc.o' failed
make[1]: *** [major_gc.o] Error 1
make[1]: Leaving directory '/home/mato/solo5/ocaml-freestanding/build/ocaml/asmrun'
Makefile:44: recipe for target 'build/ocaml/asmrun/libasmrun.a' failed
make: *** [build/ocaml/asmrun/libasmrun.a] Error 2
````
Looks like the newer openlibm moved `openlibm.h` from `src` to `include`, however just changing that path in the top-level (ocaml-freestanding) Makefile is not enough -- it then fails while looking for `cdefs-compat.h`. So needs more testing, not sure how/why it worked for you.

@talex5 Using ocaml-freestanding and friends for the Xen backend is part of the long-term plan, however, someone needs to do the work. The main issue I can anticipate is that a `nolibc/sysdeps_minios.c` may be problematic due to conflicts with Mini-OS headers...

mato:
@hannesm Ah, I see now why the test may have "succeeded" for you. `make clean` did not remove the downloaded `openlibm.tar.gz`, so you probably kept using the old one throughout. Fixed in 0b68c80. I'm wondering whether or not it'd be cleaner to use a git submodule for the openlibm tree...?

hannesm:
ups, I guess I didn't do sufficient testing. sorry.

while downloading during build is a bad idea, I always struggle with git submodules (well, in this case it makes sense).

hannesm:
@mato I fixed this PR (needed `-I openlibm/include` in `CFLAGS`)

mato:
Tested, LGTM. Bit of a shame that we have to install openlibm private headers from src/ to get openlibm_fenv.h, but not much we can do about that.

PR #5, opened 2016-07-21
FreeBSD still has amd64

hannesm:


mato:
LGTM. You'll want to make the same change to the OPAM files in solo5/solo5.

hannesm:
@mato yes.  I struggle a bit with which repos should contain what?  ATM in order to use solo5, there's either need for solo5/opam-solo5 or mirage/mirage-dev.  We should remove packages which are upstream now.  Should solo5/opam-solo5 point to mirage master (atm it uses djwillia's fork) - but if it does, there are dependencies to be dragged in (like tcpip master etc.)?

or should solo5-opam be merged into mirage-dev (it is partially)?

mato:
@hannesm I think we should merge the development packages in opam-solo5 into mirage-dev and deprecate/remove the former (after confirming with @djwillia when he's back).

PR #8, opened 2016-10-12
Support more compiler versions

mato:
Support OCaml up to `4.04.*+*`.

Also reinstate -fno-strict-aliasing and -fwrapv to OCAML_CFLAGS.


configure.sh
```diff
@@ -21,9 +21,12 @@ fi
 FREESTANDING_CFLAGS="$(pkg-config --cflags ${PKG_CONFIG_DEPS})"
 
 case $(ocamlopt -version) in
-    4.02.3)
+    4.02.[3-9])
```
hannesm: I think 4.02.3 will be the last release of the 4.02 series, no need for the complexity here

configure.sh
```diff
@@ -21,9 +21,12 @@ fi
...
         ;;
-    4.03.0)
+    4.03.[0-9])
```
hannesm: same here, there won't be a 4.03.X with X != 0.

mato:
Waiting to merge this until I can do an end-to-end test with mirage on 4.04.

mato:
This will have to wait until at least janestreet/ppx_optcomp#2 lands, so leaving unmerged for now.

hannesm:
@mato which package on Mirage side depends on `ppx_optcomp`?

mato:
@hannesm Just been through this with @yomimono, `ipaddr` is the culprit.

mato:
Rebased and re-tested against 4.04.0 release, MirageOS HTTP unikernel running on top happily serves 100k+ requests, so merging.

PR #9, opened 2016-10-14
FreeBSD/clang support

mato:
- Add FreeBSD-specific Makefile for building libasmrun.a. This is
currently identical to the Linux one except for replacing 'gcc' with 'cc'.
- Remove global -Wno-maybe-uninitialized as it's unsupported by clang,
replace with GCC-specific pragma in nolibc/dtoa.c.
- Remove nolibc sys/cdefs.h. This used to be required by openlibm, with
the move to upstream is no longer needed and conflicts with
FreeBSD/clang headers.


hannesm:
I confirm this works on my -CURRENT system!

PR #11, opened 2016-11-09
Use git subtree for OpenLibm

mato:
Rather than a (fragile) download of openlibm, use `git subtree` and keep a local copy.

For reference, the magic command is:

    git subtree add --prefix openlibm https://github.com/JuliaLang/openlibm.git v0.5.4 --squash

While at it, update OpenLibm from 0.5.1 to latest release 0.5.4.

@hannesm We discussed removing the download-during-build some time ago, are you OK with this approach?


hannesm:
LGTM.  it would be nice to document (somewhere in the README maybe) how to potentially update to newer openlibm releases (is it simply running the `git subtree add` mentioned above? -- best to have it in the README because this PR will at some point be closed and it is pretty hard to find information on GitHub once closed...)

mato:
@hannesm Done

PR #12, opened 2016-11-10
no str stubs anymore

hannesm:
The Str module of the OCaml stdlib is a pretty bad (imperative, stateful)
regular expression implementation using C code which is not very nice.

There are better solutions (like re if you really need regular expressions, or
astring if you only need basic string splitting etc.).


mato:
Fine with me, merge whenever you feel you have enough agreement in mirage/mirage#663.

PR #13, opened 2016-11-13
pkg-config files now installed into share, rather than lib

hannesm:


mato:
Closing due to different solution used

PR #14, opened 2017-01-18
declare to be freestanding

hannesm:
see https://github.com/mirage/mirage-platform/pull/178 and https://github.com/mirage/mirage/issues/660 - would be great to have this included before a release (note: i haven't tested this change) -- sorry @mato, I see you already did a release...

### DIFF
--- a/config/s.h
+++ b/config/s.h
@@ -1,4 +1,3 @@
-#define OCAML_OS_TYPE "Unix"
 #define OCAML_STDLIB_DIR "/usr/local/lib/ocaml"
 #define POSIX_SIGNALS
 //#define HAS_GETRUSAGE
--- a/configure.sh
+++ b/configure.sh
@@ -22,12 +22,15 @@ FREESTANDING_CFLAGS="$(pkg-config --cflags ${PKG_CONFIG_DEPS})"
 
 case $(ocamlopt -version) in
     4.02.3)
+        echo '#define OCAML_OS_TYPE "Unix"' >> config/s.h
         ;;
     4.03.0)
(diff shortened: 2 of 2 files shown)


mato:
Can you provide a unikernel to test this with? I can make an 0.2.1 release easily enough...

hannesm:
I'm just building mirage-skeleton for virtio with this change.. let you
know about the results

hannesm:
(this PR won't work with 4.03, the change has to be guarded (and only applied to 4.04+))

hannesm:
updated the patch (moved to `configure.sh`, and only set os_type to freestanding on 4.04+.

Testing unikernel (derived from console):

config.ml:
```OCaml
open Mirage

let main =
  foreign
    ~packages:[package "duration"]
    "Unikernel.Main" (console @-> time @-> job)

let () =
  register "console" [main $ default_console $ default_time ]
```

unikernel.ml
```OCaml
open Lwt.Infix

module Main (C: Mirage_types_lwt.CONSOLE) (Time: Mirage_types_lwt.TIME) = struct

  let start c _time =
    let rec loop = function
      | 0 -> Lwt.return_unit
      | n ->
        C.log c ("hello on " ^ Sys.os_type) >>= fun () ->
        Time.sleep_ns (Duration.of_sec 1) >>= fun () ->
        C.log c "world" >>= fun () ->
        loop (pred n)
    in
    loop 4

end
```

mato:
LGTM, waiting for Travis.

mato:
Have cut a v0.2.1 release with this change and updated the mirageos-3-beta PR.

PR #20, opened 2017-07-13
OCaml 4.05.0 support: the new afl.c needs sys/shm.h (and sys/ipc.h as…

hannesm:
… dependency), and various symbols

PR #23, opened 2017-07-14
Support OCaml 4.05

mato:
Straightforward apart from hack in configure.sh to disable the code in
afl.c by defining `__ANDROID__` for this module only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,10 @@ install: wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/ma
 script: bash -ex .travis-opam.sh
 env:
     global:
-        - EXTRA_REMOTES="https://github.com/mirage/mirage-dev.git"
         - TESTS=false
     matrix:
+        - OCAML_VERSION=4.05 EXTRA_DEPS="solo5-kernel-ukvm"
+        - OCAML_VERSION=4.05 EXTRA_DEPS="solo5-kernel-virtio"
         - OCAML_VERSION=4.04 EXTRA_DEPS="solo5-kernel-ukvm"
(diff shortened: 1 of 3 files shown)


mato:
@hannesm OK with this change as a more minimal variation on #20?

mato:
Travis seems happy so I'm going to merge this rather than #20, so as to get a release out ASAP.

avsm:
I prefer the change of defining __ANDROID__ rather than stubbing out shm.  I think this needs improving in upstream OCaml as well -- the afl.c should be using a `HAS_SHM` feature check rather than the Windows/Android check it currently has. cc @stedolan

mato:
I've verified that the mirage-skeleton `tutorial/hello` and `device-usage/network` build and run OK with this change (using OCaml 4.05.0 and solo5-kernel-ukvm).

Note that there seem to be other problems further down the Mirage stack that may be 4.05 related as a HTTP unikernel does not build for me (fails to install dependencies at mirage configure time), but that is unrelated to this change.

hannesm:
thx @mato -- sorry was travelling and couldn't respond earlier

hannesm:
just FYI for 4.06 we won't need the `-D__ANDROID__` anymore, since https://github.com/ocaml/ocaml/pull/1278 got merged

mato:
@hannesm Is that likely to also land in a 4.05.x point release?

hannesm:
@mato I don't know (neither whether a 4.05.x is planned, neither how to propose merging changes into the 4.05.releng branch)

hannesm:
@mato but we should be future-proof since we don't run configure and don't define `HAS_SYS_SHM_H` in our `s.h`.

avsm:
There's no 4.05.1 point release planned at present, so this will just be in 4.06.0

PR #25, opened 2017-07-19
aarch64 support

### LINKED ISSUE #15: ARM64 support
This issue is for discussing how ARM64 should be implemented in ocaml-freestanding. It is a counterpart to the Solo5 issue at solo5/solo5#151.

For ocaml-freestanding, the changes should be much simpler. Roughly:

* `configure.sh` needs to be taught about ARM64, and the corresponding OCaml `configure` products need to be added to `config/`.
* `nolibc` needs ARM64 support. ~~This is probably best done by pulling in the existing infrastructure from musl, i.e. adding the appropriate subset of `arch/arm/bits/*`, `arch/x86_64/bits/*` and integrating these into the build system.~~ For our purposes, this is just a few extra `#define`s added to `include/`.

mato:
See individual commits for details. Fixes #15.


mato:
Depends on Solo5/solo5#193, cc Solo5/solo5#151.

Makefile
```diff
@@ -27,8 +29,8 @@ build/ocaml/Makefile:
 
 build/ocaml/config/Makefile: build/ocaml/Makefile
 	cp config/s.h build/ocaml/config/s.h
-	cp config/m.x86_64.h build/ocaml/config/m.h
-	cp config/Makefile.$(shell uname -s).x86_64 build/ocaml/config/Makefile
+	cp config/m.$(ARCH).h build/ocaml/config/m.h
+	cp config/Makefile.$(OS).$(ARCH) build/ocaml/config/Makefile
```
hannesm: this is fine, but FreeBSD uses `amd64` as arch (`uname -m` returns `amd64`) -- we'll need to move `Makefile.FreeBSD.x86_64 -> Makefile.FreeBSD.amd64`

configure.sh
```diff
@@ -38,15 +50,15 @@ case $(ocamlopt -version) in
     4.04.[1-9]|4.04.[1-9]+*)
         OCAML_EXTRA_DEPS=build/ocaml/byterun/caml/version.h
         echo '#define OCAML_OS_TYPE "freestanding"' >> config/s.h
-        echo '#define INT64_LITERAL(s) s ## LL' >> config/m.x86_64.h
+        echo '#define INT64_LITERAL(s) s ## LL' >> config/m.${ARCH}.h
```
hannesm: see above, on FreeBSD `x86_64` is `amd64`.

Makefile
mato: Addressed in fec9f54, take a look and ack if OK.

hannesm:
LGTM, I'll test tonight and report back if it works for me.

hannesm:
this works fine on my FreeBSD system! :)

mato:
Rebased onto master and updated openlibm to master for aarch64 changes.

PR #26, opened 2017-09-23
provide uint_max, shrt_max, shrt_min, and ushrt_max, needed by newer gmp

hannesm:
my initial goal was to upgrade zarith-freestanding to the latest zarith release (1.6).

while digging down the rabbit hole, I discovered that we are still using a gmp 6.0.0a, where the newest is 6.1.2 which has several issues fixed (excerpt from https://gmplib.org/gmp6.1.html):
- Work around faulty cpuid on some recent Intel chips (this allows GMP to run on Skylake Pentiums). 
- Speedup for Intel Broadwell and Skylake through assembly code making use of new ADX instructions.

Turns out, newer gmp needs some defines: `UINT_MAX`, `SHRT_MAX`, `SHRT_MIN`, `USHRT_MAX`; which are added in this PR.  Once this is fine to merge, we'll need a new ocaml-freestanding release to do a new gmp-freestanding release in opam-repository.
Newer gmp additionally need `strstr`, which I added from musl.

### DIFF
--- a/nolibc/include/limits.h
+++ b/nolibc/include/limits.h
@@ -1,8 +1,12 @@
 #ifndef _LIMITS_H
 #define _LIMITS_H
 
-#define INT_MAX  0x7fffffff
-#define INT_MIN  (-1-0x7fffffff)
+#define INT_MAX   0x7fffffff
+#define INT_MIN   (-1-0x7fffffff)
+#define UINT_MAX  0xffffffffU
+#define SHRT_MAX  0x7fff
+#define SHRT_MIN  (-1-0x7fff)
+#define USHRT_MAX 0xffff
 #if defined(__x86_64__) || defined(__aarch64__)
 #define LONG_MAX  0x7fffffffffffffffL
 #define LLONG_MAX  0x7fffffffffffffffLL
--- a/LICENSE
+++ b/LICENSE
@@ -23,8 +23,8 @@ SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 The files ctype.c, fprintf.c, memchr.c, memcmp.c, memcpy.c, memmove.c,
 memset.c, printf.c, snprintf.c, stpncpy.c, strchr.c, strchrnul.c, strcmp.c,
-strncpy.c, strlen.c, vfprintf.c, vsnprintf.c in the nolibc/ directory are
-derived from musl libc, licensed under the following standard MIT license:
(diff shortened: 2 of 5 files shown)


hannesm:
A opam repository for testing purposes (which includes gmp-freestanding.6.1.2 and zarith-freestanding.0.6) is available at https://github.com/hannesm/opam-repository/tree/zarith-freestanding

nolibc/include/limits.h
```diff
@@ -1,8 +1,12 @@
...
 #define _LIMITS_H
 
-#define INT_MAX  0x7fffffff
-#define INT_MIN  (-1-0x7fffffff)
+#define INT_MAX   0x7fffffff
+#define INT_MIN   (-1-INT_MAX)
+#define UINT_MAX  (2U*INT_MAX+1)
+#define SHRT_MAX  0x7fff
+#define SHRT_MIN  (-1-SHRT_MAX)
+#define USHRT_MAX (2*SHRT_MAX+1)
```
mato: Can you use the same definitions (verbatim) as those used in the original musl [limits.h](http://git.musl-libc.org/cgit/musl/tree/include/limits.h)? That allows for easier eyeballing that the constants are indeed the same.

mato:
> A opam repository for testing purposes (which includes gmp-freestanding.6.1.2 and zarith-freestanding.0.6) is available at https://github.com/hannesm/opam-repository/tree/zarith-freestanding

I presume you meant zarith-freestanding.1.6? Just checking, so, we'd do a new release of ocaml-freestanding with this change and then add a dependency on >= (NEW_VERSION) to the new gmp-freestanding, right?

Other than the minor comment about the constant definitions not being exactly the same as musl's, this looks fine to me.

nolibc/include/limits.h
hannesm: sure, done in df9c32dbc1443f4a65f7a242143fd09b97ce0d7b - sorry for that.

hannesm:
yes, I meant zarith-freestanding.1.6.  and yes, we'll need to merge this PR, do a release of ocaml-freestanding, and then depend in gmp-freestanding on the new ocaml-freestanding.

I needed to add `--with-pic` on my FreeBSD system to get gmp-freestanding.6.1.2 and zarith.1.6 to work.

mato:
Thanks @hannesm, this LGTM so merging to master, release will have to wait for a bit longer...

PR #27, opened 2017-11-16
Support OCaml 4.06.0

mato:
Due to the refactoring of the compiler build system in 4.06.0 this is not the usual trivial change, and now needs separate paths for >= 4.06.0. I'm not entirely happy with the hacks involved but "it works for me".

Tested on Linux with ukvm and the mirage-skeleton `device-usage/network` example; I'd appreciate more testing and review from others before merging this (/cc @hannesm).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ env:
     global:
         - TESTS=false
     matrix:
+        - OCAML_VERSION=4.06 EXTRA_DEPS="solo5-kernel-ukvm"
+        - OCAML_VERSION=4.06 EXTRA_DEPS="solo5-kernel-virtio"
         - OCAML_VERSION=4.05 EXTRA_DEPS="solo5-kernel-ukvm"
         - OCAML_VERSION=4.05 EXTRA_DEPS="solo5-kernel-virtio"
         - OCAML_VERSION=4.04 EXTRA_DEPS="solo5-kernel-ukvm"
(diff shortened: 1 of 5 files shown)


hannesm:
@mato thanks for doing this.

We now need separate paths (for `<4.06` and `>= 4.06`) now, but do we really need the bits from the bigarray subdirectory? https://github.com/ocaml/ocaml/pull/1077 is where most symbols got moved into the OCaml runtime.

The header file dance is due to https://github.com/ocaml/ocaml/pull/1127

I'll test this tomorrow and report back!

hannesm:
I tested this as is on FreeBSD, and it works nicely!

mato:
@hannesm Thanks for the pointers to the OCaml issues. Based on my reading of the comments there it looks like my approach is correct. I've pushed a small change that improves the naming for the variable triggering the >=4.06.0 build paths, otherwise I think this is fine.

Regarding the continued need for `libotherlibs.a` post 4.06.0, I don't know enough about whether or not those symbols are used further up the stack. At this point I'd just merge as-is since it won't hurt, and anyone who wants to investigate further can do that in a followup PR.

hannesm:
@mato yes, fully agreed.  and to me it looks like at some point in 4.07+, OCaml may include the entire bigarray (apart from the map_file) into its stdlib.  once this happens, we can get rid of otherlibs, but with 4.06 we need it.

Issue #16, opened 2017-04-20
string_of_float returns `-NaN` for most numbers formatted by printf


timbertson:
I found this issue where yojson objects with timestamps (floats) were being serialized OK, but when deserialized they would came back as `-NaN`.

To test, I ran the following code:

```
    let float_test f : unit =
      let s = string_of_float f in
      let s_printf = Printf.sprintf "%f" f in
      Printf.printf "float=%f -> string_of_float=%s -> parsed=%f\n" f s (float_of_string s);
      Printf.printf "float=%f -> printf-string  =%s -> parsed=%f\n" f s_printf (float_of_string s_printf)
    in

    float_test 2.4;
    float_test 1493895831.871153;
    let attempts = ref 500 in
    while !attempts > 0 do
      decr attempts;
      let f = Random.float 999999999999999.0 in
      float_test f;
      float_test f
    done;
```

I got the following output in `mirage-ukvm`, with ocaml-freestanding 0.2.1:

```
float=2.400000 -> string_of_float=2.4 -> parsed=2.400000
float=2.400000 -> printf-string  =2.400000 -> parsed=2.400000
float=1493895831.871153 -> string_of_float=1493895831.87 -> parsed=1493895831.870000
float=1493895831.871153 -> printf-string  =1493895831.871153 -> parsed=-nan
float=481705019586434.312500 -> string_of_float=4.81705019586e+14 -> parsed=481705019586000.000000
float=481705019586434.312500 -> printf-string  =481705019586434.312500 -> parsed=-nan
float=481705019586434.312500 -> string_of_float=4.81705019586e+14 -> parsed=481705019586000.000000
float=481705019586434.312500 -> printf-string  =481705019586434.312500 -> parsed=-nan
float=554380010410619.812500 -> string_of_float=5.54380010411e+14 -> parsed=554380010411000.000000
float=554380010410619.812500 -> printf-string  =554380010410619.812500 -> parsed=-nan
float=554380010410619.812500 -> string_of_float=5.54380010411e+14 -> parsed=554380010411000.000000
float=554380010410619.812500 -> printf-string  =554380010410619.812500 -> parsed=-nan
```

i.e. when using `string_of_float` the number correctly roundtrips to a string and back, but using `printf`'s float formatting it consistently gets parsed as `-NaN`.
`2.4` was a random simple case I tried, and funnily enough it works both ways - perhaps the bug only affects numbers with a large number of decimal places?

[...]

PR #28, opened 2017-11-21
Update dtoa to latest version

mato:
Update nolibc/dtoa.c to the latest version at http://www.netlib.org/fp/, with the following modifications:

- Perform arch-specific configuration inline in dtoa.c.
- Add and use a STRTOD_ONLY flag to compile only strtod() and its dependencies.
- Make gethex() static.

Further, nolibc/dtoa.c is compiled with -fno-strict-aliasing. This is strictly necessary only for clang, but not worth the extra detection infrastructure so we do it unconditionally for all compilers. For details, see https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=216770.

### DIFF
--- a/nolibc/Makefile
+++ b/nolibc/Makefile
@@ -17,6 +17,8 @@ OBJS=ctype.o \
      stubs.o \
      vfprintf.o vsnprintf.o snprintf.o fprintf.o printf.o
 
+dtoa.o: CFLAGS+=-fno-strict-aliasing
+
 libnolibc.a: $(OBJS) $(SYSDEP_OBJS)
 	$(AR) rcs $@ $(OBJS) $(SYSDEP_OBJS)
 
--- a/nolibc/dtoa.c
+++ b/nolibc/dtoa.c
@@ -17,6 +17,21 @@
  *
  ***************************************************************/
 
+/* Configuration and modifications specific to compliation for
+ * ocaml-freestanding.
+ */
+
+#if defined(__x86_64__) || defined(__aarch64__)
+#define IEEE_8087
+#define Long int
+#else
+#error Unsupported architecture
+#endif
+
+/* Compile only strtod() and its dependencies
+ */
+#define STRTOD_ONLY
+
 #if !defined(__clang__)
 #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
 #endif
@@ -107,7 +122,6 @@
  *	something other than "long long", #define Llong to be the name,
  *	and if "unsigned Llong" does not work as an unsigned version of
  *	Llong, #define #ULLong to be the corresponding unsigned type.
- * #define KR_headers for old-style C function headers.
  * #define Bad_float_h if your system lacks a float.h or if it does not
  *	define some or all of DBL_DIG, DBL_MAX_10_EXP, DBL_MAX_EXP,
  *	FLT_RADIX, FLT_ROUNDS, and DBL_MAX.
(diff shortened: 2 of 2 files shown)


mato:
@hannesm Could you give this a test on FreeBSD? It'll be quicker than me doing it.
@timbertson This fixes the C test case in #16 for me, can you confirm that it works for you?

hannesm:
@mato thanks, this compiles fine with `FreeBSD clang version 5.0.0 (branches/release_50 310316) (based on LLVM 5.0.0svn)`, but fails to link a unikernel:
```
/home/hannes/.opam/ino64/lib/pkgconfig/../../lib/ocaml-freestanding/libnolibc.a(dtoa.o): In function `strtod':
dtoa.c:(.text+0x16c5): undefined reference to `__flt_rounds'
dtoa.c:(.text+0x28fa): undefined reference to `__flt_rounds'
```

hannesm:
in FreeBSD, `__flt_rounds` is defined in `lib/libc/amd64/gen/flt_rounds.c` as:
```
/*
 * Written by J.T. Conklin, Apr 10, 1995
 * Public domain.
 */

#include <sys/cdefs.h>
__FBSDID("$FreeBSD$");

#include <float.h>

static const int map[] = {
        1,      /* round to nearest */
        3,      /* round to zero */
        2,      /* round to negative infinity */
        0       /* round to positive infinity */
};

int
__flt_rounds(void)
{
        int x;

        /* Assume that the x87 and the SSE unit agree on the rounding mode. */
        __asm("fnstcw %0" : "=m" (x));
        return (map[(x >> 10) & 0x03]);
}
```

mato:
@hannesm I don't understand what code in `dtoa.c` is generating that call. AFAICT, it should only call `fegetround()` if `Honor_FLT_ROUNDS` is defined, which it's not.

hannesm:
@mato solo5 copies `x86/float.h` which includes a `#define FLT_ROUNDS      __flt_rounds()`, and `FLT_ROUNDS` is used in `dtoa.c` (`#define Flt_Rounds FLT_ROUNDS`, where `Flt_rounds` is used in the code, and another `#define Rounding Flt_Rounds` where `Rounding` is used quite a lot in the code).

mato:
@hannesm I've pushed a hacky fix for this in a6eb64a which reverts to the previously used value and does not depend on FLT_ROUNDS. Let me know if this fixes the linking on FreeBSD.

hannesm:
@mato yes, this links now (tested with `tutorial/hello` from `mirage-skeleton`).  is there a test unikernel for `dtoa`?  I couldn't find any in `mirage-skeleton` - might be worth to add one!?

mato:
@hannesm You can use @timbertson's code from https://github.com/mirage/ocaml-freestanding/issues/16#issue-223060795 to test, though that only tests `string_of_float`. I've not looked further to see what (if any) other code paths might use `strtod()` in the OCaml runtime. I've tried running that code against this version and it works as expected (no more `-NaN`).

Do you think this is OK to merge? It should make anything worse, fingers crossed, but then I'm not an expert on the floating point setup...

hannesm:
@mato thanks again, the output looks good:
```
float=2.400000 -> string_of_float=2.4 -> parsed=2.400000
float=2.400000 -> printf-string  =2.400000 -> parsed=2.400000
float=1493895831.871153 -> string_of_float=1493895831.87 -> parsed=1493895831.870000
float=1493895831.871153 -> printf-string  =1493895831.871153 -> parsed=1493895831.871153
float=926038888360970.250000 -> string_of_float=9.26038888361e+14 -> parsed=926038888361000.000000
float=926038888360970.250000 -> printf-string  =926038888360970.250000 -> parsed=926038888360970.250000
float=926038888360970.250000 -> string_of_float=9.26038888361e+14 -> parsed=926038888361000.000000
float=926038888360970.250000 -> printf-string  =926038888360970.250000 -> parsed=926038888360970.250000
```

IMHO good to merge and it improves the current situation!

mato:
I've also run Tim's test code and the output is fine, but will leave this a while to see if he can confirm that the original issue is fixed. If I don't get a response within a day or two then I'll merge it.

timbertson:
Thanks @mato. I'm afraid I don't have time (or the right environment) to test this anytime soon, but it looks like it's working for you both, so I'd say go ahead 👍

mato:
@timbertson Thanks for the feedback.

I've verified that GCC on Linux uses a fixed value of `1` for `FLT_ROUNDS` in `float.h`, have updated the last commit to reflect that comment and will merge shortly.

PR #29, opened 2017-12-07
add dlmalloc to ocaml-freestanding (formerly in solo5)

djwillia:
This is part of https://github.com/Solo5/solo5/issues/223

Malloc functions are now implemented by dlmalloc in malloc.c.  The `sbrk` function is implemented in sysdeps_solo5.c and keeps track of the raw memory that it can give to malloc for "morecore".  `sbrk` gets the memory information from the new `solo5_get_info` interface (see https://github.com/Solo5/solo5/pull/224)

### DIFF
--- a/nolibc/sysdeps_solo5.c
+++ b/nolibc/sysdeps_solo5.c
@@ -70,20 +70,30 @@ clock_t times(struct tms *buf)
 }
 
 
-static struct solo5_info info;
+static struct solo5_mem_info info;
 /*
  * Called by dlmalloc to allocate or free memory.
  */
 void *sbrk(intptr_t increment)
 {
-    static uint64_t heap_top;
-    uint64_t prev, brk;
+    static uint64_t heap_top, stack_guard_size;
 
-    if (!heap_top) {
-        solo5_get_info(&info);
+    /* One-time initialization. */
+    if ((!heap_top) || (!stack_guard_size)) {
+        solo5_mem_info(&info);
         heap_top = info.heap_start;
+
+        /*
+         * If we have <1MB of free memory then don't let the heap grow
+         * to more than roughly half of free memory, otherwise don't
+         * let it grow to within 1MB of the stack.
+         */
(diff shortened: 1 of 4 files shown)


nolibc/sysdeps_solo5.c
```diff
@@ -91,3 +68,32 @@ clock_t times(struct tms *buf)
...
+/*
+ * Called by dlmalloc to allocate or free memory.
+ */
+void *sbrk(intptr_t increment)
+{
+    static uint64_t heap_top;
+    uint64_t prev, brk;
+
+    if (!heap_top) {
+        solo5_get_info(&info);
```
ricarkol: This is what would go to a ocaml_freestanding Init. I remember you mentioned it; you must have thought of a reason why to not do that instead.

djwillia: I was going to move the application entry point out of mirage-solo5, but found it was going to be a little bit messy with things like how to pass the command line around.  Also, this way we don't need to modify mirage-solo5.

djwillia:
It looks like travis won't work because it would need to pin a different version of solo5 that accompanies this PR (https://github.com/Solo5/solo5/pull/224).  Is there a way to do that for travis tests?  @mato

nolibc/sysdeps_solo5.c
ricarkol: Sounds good, better not to touch mirage-solo5 then.

hannesm:
@djwillia I'm not mato, but in `.travis.yml` you can add to the `env -> global` section where `TESTS=false` lives `PINS="solo5-kernel-ukvm:https://github.com/Solo5/solo5.git#mybranch"` - where `PINS` is documented [here](https://github.com/ocaml/ocaml-ci-scripts/blob/master/README-travis.md#customizing-the-opam-pin-set) as a whitespace-separated list of `<package>:<source_or_version>`.

mato:
Superseded by #32.

PR #31, opened 2018-02-09
opam, Travis CI: Switch to Solo5 master for development

mato:
Now that Solo5/opam-solo5 is back online for development on master,
switch ocaml-freestanding to it by adding EXTRA_REMOTES to Travis CI and
updating OPAM metadata to depend on Solo5 >= 0.3.0.

As part of this, based on the earlier consensus on mirage-dev regarding
supported compiler versions and to speed up Travis CI cycles, remove
compilers older than 4.04.2 from the build matrix and update
ocaml-version in opam.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ script: bash -ex .travis-opam.sh
 env:
     global:
         - TESTS=false
+        - EXTRA_REMOTES="git://github.com/Solo5/opam-solo5"
     matrix:
         - OCAML_VERSION=4.06 EXTRA_DEPS="solo5-kernel-ukvm"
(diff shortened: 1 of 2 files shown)


mato:
I have no idea what the problem is here. Either Travis is confused, or there is something about the package versions in https://github.com/solo5/opam-solo5 that it doesn't like, but I can't figure out what...

hannesm:
@mato to me, it looks like only the first travis job re-ran after your rename `3.0.0~dev` to `solo5~dev`.  I just restarted the remaining 5 travis jobs.

hannesm:
(travis is green now!)

PR #33, opened 2018-02-20
cleanup README.md, configure.sh, and opam to meet on OCaml versions

hannesm:
since we dropped <4.04.2 in opam, we should also drop these versions from README and configure


mato:
Seems reasonable. No one has complained on the mirageos mailing list either about the >=4.04.2 constraint, so let's do it. Will merge after the failing Travis (which I restarted) builds. Thanks.

PR #35, opened 2018-04-08
Add OpenBSD support

adamsteen:
add OpenBSD support
update INT64 and UNIT64 as per hannesm suggestion in config.in/m.x86_64.h
add OPAM 2 support


config.in/m.x86_64.h
```diff
@@ -3,8 +3,8 @@
 #define SIZEOF_LONG 8
 #define SIZEOF_PTR 8
 #define SIZEOF_SHORT 2
-#define ARCH_INT64_TYPE long
-#define ARCH_UINT64_TYPE unsigned long
+#define ARCH_INT64_TYPE long long
+#define ARCH_UINT64_TYPE unsigned long long
```
hannesm: this breaks on Linux (see travis CI), we'll need to make this conditional depending on the OS (similar to s.h (append in configure.sh), or Makefile.$OS.h)...

adamsteen: something like the following in configure.sh? and remove them from `config/m.x86_64.h`

```
if [ "${BUILD_ARCH}" = "x86_64" ]; then
    if [ "${BUILD_OS}" = "OpenBSD" ]; then
        OPENLIBM_COMPILER="USECLANG=1"
        echo '#define ARCH_INT64_TYPE long long' >> config/m.x86_64.h
        echo '#define ARCH_UINT64_TYPE unsigned long long' >> config/m.x86_64.h
    else
        echo '#define ARCH_INT64_TYPE long' >> config/m.x86_64.h
        echo '#define ARCH_UINT64_TYPE unsigned long' >> config/m.x86_64.h
    fi
fi
```

that doesn't feel right, but i am not sure how else to do it

opam
```diff
@@ -21,5 +21,5 @@ conflicts: [
 available: [
   ocaml-version >= "4.04.2" & ocaml-version < "4.07.0" &
   ((os = "linux" & (arch = "x86_64" | arch = "aarch64") |
-   (os = "freebsd" & arch = "amd64")))
+   (os = "freebsd" & arch = "amd64") | (os = "openbsd" & (arch = "amd64" | arch = "x86_64"))))
```
mato: Why do you need both `amd64` and `x86_64` here?

configure.sh
```diff
@@ -20,6 +20,7 @@ if [ -z "${PKG_CONFIG_DEPS}" ]; then
 fi
 ocamlfind query ocaml-src >/dev/null || exit 1
 
+OPENLIBM_COMPILER=
```
mato: Why is this (and the associated `Makefile` change) required on OpenBSD but not on FreeBSD which also uses clang? Can you achieve the same by just passing through `CC` to the openlibm Makefile?

config.in/m.x86_64.h
mato: A better way would be to just put the OpenBSD-specific types inside an `#if defined(__OpenBSD__)` block directly in `m.x86_64.h`. This moves the decision making to the single place where it's required.

opam
adamsteen: this was a fix for OPAM 2, i will remove and eventually get OPAM 2 updated.

configure.sh
adamsteen: [openlibm - Fixes for building on OpenBSD #176](https://github.com/JuliaLang/openlibm/pull/176)

Fixed in openlibm 5 days ago.

adamsteen: openlibm defaults to gcc without `OPENLIBM_COMPILER="USECLANG=1"` on OpenBSD till the build include openlibm PR 176.

```
gcc: unrecognized option '-nostdlibinc'
cc1: error: -fno-gnu89-inline is not supported
gmake[1]: *** [Make.inc:141: src/common.c.o] Error 1
gmake: *** [Makefile:22: build/openlibm/libopenlibm.a] Error 2  
```

Passing CC to the Makefile wont help as it overrides it based on USECLANG or USEGCC [Make.inc](https://github.com/JuliaLang/openlibm/blob/master/Make.inc)

mato: Ah, OK. In that case I'll bump the version of openlibm used by ocaml-freestanding to include that change so that we don't need the extra hacks.

opam
mato: This seems related to #36, it looks like there has been a change in the way OPAM 2 canonicalises the host arch. Am investigating.

mato:
@adamsteen I've updated openlibm in #37, could you please rebase this on latest master, remove the `OPENLIBM_COMPILER` change and squash your commits? Then we should be good to go!

adamsteen:
@mato Its all ready to go!

mato:
Tested with today's build of OPAM 2 on OpenBSD 6.3, nothing controversial here, so merging.

PR #36, opened 2018-04-11
Fix opam arch constraint

kit-ty-kate:
```arm64``` is the correct arch name not ```aarch64```

### DIFF
--- a/opam
+++ b/opam
@@ -20,6 +20,6 @@ conflicts: [
 ]
 available: [
   ocaml-version >= "4.04.2" & ocaml-version < "4.07.0" &
-  ((os = "linux" & (arch = "x86_64" | arch = "aarch64") |
+  ((os = "linux" & (arch = "x86_64" | arch = "arm64") |
    (os = "freebsd" & arch = "amd64")))
 ]


kit-ty-kate:
arf, maybe, sorry. I haven't tested with opam1 :/ @AltGr ?

kit-ty-kate:
Right, the documentation of opam1 says: ```arch: the host architecture, as returned by uname -m```
but opam2 says something more disturbing: ```arch: the host architecture, typically one of "x86_32", "x86_64", "ppc32", "ppc64", "arm32" or "arm64", or the lowercased output of uname -m, or "unknown"```

I don't understand why ```arm64``` is used as an example when uname -m returns ```aarch64```. On the arm64 machine I have access to only ```arm64``` works (with opam2)

mato:
I can't check right now, but will take a look next week at what exactly OPAM 2 returns on my ARM64 board(s). In the mean time, it'd be helpful to see if we can find/document the exact OPAM 2 changes that lead to this, as it looks like it may lead to even more divergence for `opam` files with `arch` constraints (see also https://github.com/mirage/ocaml-freestanding/pull/36).

/cc @avsm @AltGr

kit-ty-kate:
I finally could check with opam1 on arm64 and only ```aarch64``` works here. So I'm going to close this PR and update the opam-repository PR.

mato:
@kit-ty-kate Which opam-repository PR? I'd still like to understand what exactly has changed in the way `arch` is set between OPAM 1 and 2.

kit-ty-kate:
The one referenced above (https://github.com/ocaml/opam-repository/pull/11768). Yes me too but for now I have my answers namely:
* On opam1: ```aarch64```
* On opam2: ```arm64```

but @AltGr can still answer more in details later even if this PR is closed.

hannesm:
FWIW @mato https://github.com/ocaml/opam/issues/3058 contains some discussion about the opam variables

PR #43, opened 2018-10-11
Heap bounds check for non-contiguous memory

ehmry:
Constrain heap allocations within the bounds of the heap specified by
solo5_start_info. This check is required for platforms that use
non-contiguous memory for the heap and stack and cannot infer the end of
heap from the current stack location.

---
This is a simple fix for the Genode bindings, which allocate and map stack and heap seperately in the unikernel address space.


mato:
Looks sensible, and will keep the current "guard" on platforms with a contiguous heap+stack area. Thank you!

PR #45, opened 2018-10-24
install an empty META file

hannesm:
when not including the META file, other build systems won't recognize ocaml-freestanding to be installed, which makes it hard to depend on ocaml-freestanding being there (e.g. as depopt). would be worth releasing a 0.4.1 with this change since e.g. bigstringaf should be fixed with the correct dependencies.


samoht:
LGTM

mato:
I'll cut a release this evening.

PR #49, opened 2019-01-31
use solo5_abort in abort

hannesm:
`solo5_abort` has been in solo5 since 0.3.0. OCaml-freestanding depends on 0.4.0 (the big rename) anyways ;)

`solo5_abort` will potentially dump a core file for later debugging. Before this PR, it was never called from the OCaml side.


avsm:
Definitely!

PR #51, opened 2019-03-16
Fix system compiler

hannesm:
as mentioned by @TheLortex in https://github.com/ocaml/opam-repository/pull/13679#issuecomment-473268721

this PR also adds a system compiler to the travis.yml matrix.


hannesm:
please note that it may be the case that there's need for an additional symbolic link, as described by @TheLortex in https://github.com/mirage/mirage-platform/pull/208#discussion_r266233750 -- I fail to find which mirage+functoria branch carries the changes to test the ocaml-freestanding release

hannesm:
(sorry, moving discussion back to the original https://github.com/mirage/mirage/issues/969)

PR #52, opened 2019-05-13
allow non-opam installations

hannesm:
Makefile: provide rules for flags/libs.tmp and flags/cflags.tmp, clean: rm them
.gitignore: ignore remporary generated flags files

this improves local non-opam builds after c5bf86c /cc @TheLortex

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,8 @@ Makeconf
 build/
 config/
 ocaml-freestanding.pc
+flags/cflags
+flags/cflags.tmp
+flags/libs
+flags/libs.tmp
+flags/libs.tmp2
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ flags/cflags
 flags/cflags.tmp
 flags/libs
 flags/libs.tmp
-flags/libs.tmp2
--- a/Makefile
+++ b/Makefile
@@ -100,6 +100,9 @@ ocaml-freestanding.pc: ocaml-freestanding.pc.in Makeconf
 	    -e 's!@@PKG_CONFIG_EXTRA_LIBS@@!$(PKG_CONFIG_EXTRA_LIBS)!' \
 	    ocaml-freestanding.pc.in > $@
 
+flags/libs.tmp:
+	opam config subst $@
+
 flags/libs: flags/libs.tmp Makeconf
 	sed -e 's!@@PKG_CONFIG_EXTRA_LIBS@@!$(PKG_CONFIG_EXTRA_LIBS)!' \
 	    flags/libs.tmp > flags/libs.tmp2
@@ -108,6 +111,9 @@ flags/libs: flags/libs.tmp Makeconf
 		env PKG_CONFIG_PATH=$(shell opam config var prefix)/lib/pkgconfig pkg-config $$PKG --libs >> flags/libs.tmp2;\
 	done
 	echo "("`cat flags/libs.tmp2`")" > flags/libs
+
+flags/cflags.tmp:
+	opam config subst $@
 
 flags/cflags: flags/cflags.tmp Makeconf
 	for PKG in $(PKG_CONFIG_DEPS); do \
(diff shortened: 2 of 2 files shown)


mato:
@hannesm I pushed 7904eee to your branch which simplifies things a bit, AFAICT the outputs are identical. FWIW I'm not too happy about the use of the

    echo "("`cat ...`")"

construct, since if there are any shell constructs in the output of `cat` then they will get evaluated by the shell, but if you're not concerned about that then I'm happy to merge this.

hannesm:
@mato thanks! I did neither write nor merge the  ``echo "("`cat ..`")"`` lines, and agree with you that it is a bad idea. I suspect they can be rewritten as:
```sh
echo "(" > flags/libs
env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
	    pkg-config $(PKG_CONFIG_DEPS) --libs  | sed -e 's!@@PKG_CONFIG_EXTRA_LIBS@@!$(PKG_CONFIG_EXTRA_LIBS)!' >> flags/libs
echo ")" >> flags/libs
```
(but likely there are some superfluous or missing line endings etc.)


my initial motivation for this PR that before it with a cloned ocaml-freestanding running `make` errored, and since I wanted to improve ocaml-freestanding (see the other PRs), this was my minimal attempt to fix it.

I guess, esp. with respect to https://github.com/mirage/mirage/issues/969#issuecomment-485593066 I'm still lacking the overall design of mirage 4, in respect to usage of dune and pkg-config.

TheLortex:
For the ``echo "("`cat ..`")"`` ugliness, that's my bad. That PR exposing flags using files was probably too early.. 

Regarding `dune` and `pkg-config`:
- `ocaml-freestanding` is currently not a `dune` project, but it's at the frontier between dune packages (like `mirage-solo5`) and non-dune packages (`solo5-bindings-...`).
- in order to gather compilation settings (cflags, link flags, ..) there are two ways:
  - Using pkg-config: this works well in most of the cases, but is difficult in an opam-less environment as there won't be an obvious pkg-config path. 
  - Using files that dune can include: this is what https://github.com/mirage/ocaml-freestanding/pull/50 enabled and it allows to get rid of pkg-config in the mirage build system. It's mainly a hack that precomputes the flags. It should work better in an opam-less environment as these flags files are accessible directly from the dune workspace (using `%{lib:ocaml-freestanding:cflags}%` to refer to the cflags for example).

It looks like we're moving toward a Mirage 4 that depends less and less on opam. According to what's suggested in https://github.com/mirage/mirage/issues/969#issuecomment-485593066, most of the sources need to be locally downloaded instead of being installed with opam.

hannesm:
IMHO: let's merge this now since it improves development of ocaml-freestanding, and adapt ocaml-freestanding piecewise with mirage4 development if needed. at least if I unerstand @TheLortex correctly, having both pkg-config and libs/cflags helps with the mirage4 development atm.

mato:
@hannesm @TheLortex Does dune interpret `\n` in the included files in any way? If not, then I have a solution for replacing the `echo` and `cat` with a single `sed` invocation.

mato:
In other words, is

    (
    foo
    bar
    baz
    )

equivalent to

    (foo bar baz)

?

hannesm:
@mato I don't know. from my experience with `dune` files, they don't care about whitespace, i.e. your above example is identical

mato:
@hannesm @TheLortex I checked with the dune metadata format doc (https://dune.readthedocs.io/en/latest/project-layout-specification.html#metadata-format) and it does seem that extra whitespace / newlines are not significant in a list, so I've pushed 0bb2f6f to remove the problematic use of `cat`.

.gitignore
```diff
@@ -2,3 +2,8 @@ Makeconf
 build/
 config/
 ocaml-freestanding.pc
+flags/cflags
+flags/cflags.tmp
+flags/libs
+flags/libs.tmp
+flags/libs.tmp2
```
hannesm: the last one is now superfluous..

mato: Fixed in eba186f (force push)

hannesm:
i approve this change as it is now.

Issue #24, opened 2017-07-14
Add some basic link tests to the build


mato:
To facilitate better detection of issues such as #21, we should consider adding some basic link tests to the build. These would verify that a full link of a (non-runnable) freestanding "Hello, World" works.

PR #53, opened 2019-05-13
do not claim to support POSIX_SIGNALS, provide alternative symbols

hannesm:
s.h: do not define POSIX_SIGNALS
signal.h: define jmp_buf, signal, setjmp
  remove sigaction/sigsetjmp/sigaddset/sigdelset/sigemptyset/sigprocmask
stubs.c: see above

The motivation for this change is that my last porting to OCaml 4.08.0 attempt required more symbols due to changes to the signal handling. with this change in, it should be easier. I tested this change with OCaml 4.07.1 and actual unikernels (hello/prng/network from mirage-skeleton), and compiled gmp-freestanding, which works flawlessly.

There is a downside, though: the binary size of libasmrun.a increased by 176 bytes. The upside is that nolibc.a reduced by 1168 bytes, so ultimately it saves around 1000 bytes _per unikernel_.

Please let me know what you think about this, to me it is an improvement (fewer symbols, the OCaml runtime code in question is never executed anyways).

### DIFF
--- a/config.in/s.h
+++ b/config.in/s.h
@@ -1,5 +1,5 @@
 #define OCAML_STDLIB_DIR "/usr/local/lib/ocaml"
-#define POSIX_SIGNALS
+//#define POSIX_SIGNALS
 //#define HAS_GETRUSAGE
 #define HAS_TIMES
 //#define HAS_SOCKETS
--- a/nolibc/include/signal.h
+++ b/nolibc/include/signal.h
@@ -1,23 +1,12 @@
 #ifndef _SIGNAL_H
 #define _SIGNAL_H
 
-typedef int sigjmp_buf;
-typedef int sigset_t;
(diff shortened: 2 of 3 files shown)


mato:
This looks fine, but I'm a bit worried that due to the absence of link-time tests for ocaml-freestanding (see #24) we might break older but still supported compiler versions. I took a stab at making some tests, but could not come up with a straightforward & clean solution.

So, if you don't mind, I'd like to ask that you manually test this with all the supported compiler compiler versions (as listed in .travis.yml). No need to test with different Solo5 bindings, one target is fine.

hannesm:
a successful travis run rebased on #55 (which links a mirageos unikernel with ocaml-freestanding) is at https://travis-ci.org/mirage/ocaml-freestanding/builds/533116887

PR #55, opened 2019-05-16
travis: add a POST_INSTALL_HOOK that links a unikernel

hannesm:
as requested in #24. From #53

> So, if you don't mind, I'd like to ask that you manually test this with all the supported compiler compiler versions (as listed in .travis.yml). No need to test with different Solo5 bindings, one target is fine.

actually, I do mind ;) -- that's why 4c1c42f includes a (basic) solution here where travis after compiling and installing ocaml-freestanding executes a shell script which installs mirage in order to compile and link a minimal unikernel.

travis run differences (IMHO acceptable for the added benefit):
- the latest release (https://travis-ci.org/mirage/ocaml-freestanding/builds/507520273) finished in 6-8.5 minutes
- the link test (https://travis-ci.org/mirage/ocaml-freestanding/builds/533096630) finishes in up to 11.5 minutes

A failing run (I removed `gettimeofday` in https://github.com/mirage/ocaml-freestanding/commit/3c2529461a3712e54d42c35dab3e47423cd804f7, required by OCaml runtime) can be seen at https://travis-ci.org/mirage/ocaml-freestanding/jobs/533099435 -- i.e. the travis run fails, and you can see the output at the end.

The second commit here, 990de6b, is an artifact of a compiler issue we encountered in mirage-kv 2.0.0 (see https://github.com/mirage/mirage-kv/commit/3df5b30f240f675d8d95c94440825e013b7cfc2f), which is not yet properly propagated through mirage opam packages (see https://github.com/mirage/mirage/pull/988). Thus it just reflects reality, and removes the 4.04.2 bits and pieces from ocaml-freestanding.


hannesm:
I've two more thoughts on this:
- A more lightweight (i.e. less dependencies) approach would be to run the link command mirage does manually in the test here, but I don't think it is worth it (given that it requires duplicating parts of mirage logic for this, and ocaml-freestanding's only user is mirage)
- Since ocaml-freestanding has two C libraries as users (namely OCaml runtime and gmp), should a second unikernel which depends on nocrypto (or zarith-freestanding) be added (or the current one modified to include such a dependency)?

mato:
@hannesm Thank you very much for this, it's a big improvement in ensuring that ocaml-freestanding does not break in unexpected ways.

> Thus it just reflects reality, and removes the 4.04.2 bits and pieces from ocaml-freestanding.

I hadn't realised that Mirage no longer supports OCaml 4.04.2, thanks for pointing this out. Nit: In the change to `README.md`, you may as well specify 4.07.**1** as the upper bound since we do test with that.

> A more lightweight (i.e. less dependencies) approach would be to run the link command mirage does manually

That was the approach I tried yesterday, but it does indeed result in a lot of duplication. Your solution using Travis `POST_INSTALL_HOOK` is much better.

> Since ocaml-freestanding has two C libraries as users (namely OCaml runtime and gmp), should a second unikernel which depends on nocrypto (or zarith-freestanding) be added [...]

This would slow down the Travis builds significantly, so I don't think it's worth it right now.

> Regarding manually specifying `MODE`

I also prefer the manual approach, but can we call it `MIRAGE_TEST_MODE` to make it more explicit?

I'm happy with merging this after addressing the minor comments (`README.md` and naming of `MODE`), followed by #53.

hannesm:
@mato done in ca675eb, I'd suggest to wait for travis and then squash-merge

PR #56, opened 2019-05-21
fixes for BSD sed compatibility

hannesm:
sorry, I wasn't carefully trying #52 on BSD.

It turns out that the expression `sed -e '1i (' ` (and `a`) does not work with BSD sed:
```
sed: 1: "1i (
": command i expects \ followed by text
```

I found some explanation [here](https://riptutorial.com/sed/topic/9436/bsd-macos-sed-vs--gnu-sed-vs--the-posix-sed-specification), using the proposed `sed -e '1 i\'$'\n''('`works in a shell, but via the Makefile leads to a leading `$` on a Linux (debian) system -- it works fine on FreeBSD :/

Instead, we can use sed replacements (and shell invocation) as done in this PR, and provide the parens in the `.in` files.

(Another solution I came up with is another conditional in Makefile:
```make
ifeq ($(BUILD_OS),Linux)
SED_POST=
else
SED_POST=\'$$'\n''
endif
```
and then `sed -e '1 i$(SED_POST)('` in the rules, but IMHO that one is less readable than the provided diff).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,22 +103,24 @@ ocaml-freestanding.pc: ocaml-freestanding.pc.in Makeconf
 flags/libs.tmp: flags/libs.tmp.in
 	opam config subst $@
 
-flags/libs: flags/libs.tmp Makeconf
+PKG_CONFIG_LIBS != \
 	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
-	    pkg-config $(PKG_CONFIG_DEPS) --libs >> $<
-	sed -e '1i (' \
-            -e 's!@@PKG_CONFIG_EXTRA_LIBS@@!$(PKG_CONFIG_EXTRA_LIBS)!' \
-	    -e '$$a )' \
+	    pkg-config $(PKG_CONFIG_DEPS) --libs
+
+flags/libs: flags/libs.tmp Makeconf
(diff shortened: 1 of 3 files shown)


mato:
Sorry, chaotic day today, I have to run and no time to open a PR, but how about this:

```
flags/libs: flags/libs.tmp Makeconf
	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
	    pkg-config $(PKG_CONFIG_DEPS) --libs >> $<
	awk -v RS= -- '{ \
	    sub("@@PKG_CONFIG_EXTRA_LIBS@@", "$(PKG_CONFIG_EXTRA_LIBS)", $$0); \
	    print "(", $$0, ")"\
	    }' $< >$@

#...

flags/cflags: flags/cflags.tmp Makeconf
	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
	    pkg-config $(PKG_CONFIG_DEPS) --cflags >> $<
	awk -v RS= -- '{ \
	    print "(", $$0, ")"\
	    }' $< >$@
```

Seems to do the trick for me, and the constructs work with both BSD awk and $RANDOM_LINUX awk. Also has the advantage of cleaning up newlines.

mato:
See #57

hannesm:
thanks for the fix!

PR #58, opened 2019-06-26
add support for OCaml 4.08.0 release

avsm:
Only compile-tested so far

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,11 @@
 
 include Makeconf
 
-ifeq ($(OCAML_GTE_4_07_0),yes)
+ifeq ($(OCAML_GTE_4_08_0),yes)
+FREESTANDING_LIBS=build/openlibm/libopenlibm.a \
+		  build/ocaml/runtime/libasmrun.a \
+		  build/nolibc/libnolibc.a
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -33,7 +37,10 @@ build/ocaml/Makefile:
 	cp -r `ocamlfind query ocaml-src` build/ocaml
 
 build/ocaml/config/Makefile: build/ocaml/Makefile
```
mato: Am I right in reading the diff, and the rule for `build/ocaml/config/Makefile:` is used for OCaml < 4.08.0, while `build/ocaml/Makefile.config:` is used for OCaml >= 4.08.0?

If that's the case, why does the former have clauses with `ifeq ($(OCAML_GTE_4_08_0),yes)` in it? I think those should be removed, and to ease future maintenance it'd also be good to explicitly mention in a comment which rule is used for what versions, and/or wrap the rule itself in the relevant `ifeq` clauses.

mato:
@avsm Thanks. I took the liberty of pushing some fixes/cleanups (e8f6113) and rebasing your branch on master. All looking good, so will merge.

PR #61, opened 2019-07-28
configure: fix 4.08+ case; support >= 4.08.1+rc3

hannesm:
I hit this case when I tried to use 4.08.1+rc2.


hannesm:
the second commit also supports >= 4.08.1+rc3 -- which expects `off_t` being specified in `sys/types.h` (which aligns with posix).

mato:
LGTM, thanks!

PR #62, opened 2019-09-19
ocaml 4.09.0 support

hannesm:
i have compiled this (it compiles), but not yet have done a full e2e test with a mirageos unikernel (gladly travis will do this for us)


mato:
I have manually tested this, just in case, the E2E test unikernel runs fine.

Issue #63, opened 2019-11-03
openbsd int64 formatter


cfcs:
https://github.com/mirage/ocaml-freestanding/blob/master/config.in/m.x86_64.h#L13

I believe this should be included in the `ifdef` guard and be set to `ll` on openbsd?
see https://man.openbsd.org/printf.3

ping @adamsteen

adamsteen:
@cfcs 

i am not sure, i haven't seen any problems using ocaml-freestanding, but if you have tested it and are happy with the results, make the change.

I am sorry i can not volunteer to test and make the fix, life circumstances have severely limited my access and time if from of a computer, but please feel free to test and make a change.

mato:
I'm not sure what is required here. There was someone on IRC reporting that "it doesn't work with OpenBSD", but did not file an issue or provide details. I do recall that OpenBSD has a different idea of integer/pointer sizes to every other *NIX on amd64.

I'd suggest running the standard OCaml `configure` on OpenBSD and comparing the resulting generated files with those in ocaml-freestanding, then reconciling any differences / adding OpenBSD-specific templates if need be.

cfcs:
I don't think this is the cause of that bug report, just noting that it is incorrect, and in theory could lead to truncation or compiler warnings/errors if that macro is used.
In theory, because it would apply on platforms where `long` and `long long` are not equivalent, which doesn't seem to be the case for any platforms we care about.
Where it could matter is if some compiler decides to get strict about types.

mato:
As far as I can tell, this is fixed by #65 which will be in the next release, so closing this issue.

PR #65, opened 2019-11-05
Build system cleanups

hannesm:
see individual commits for details if interested (but first take a look whether CI is happy). comments welcome if there are controversial changes (such as calling OCaml's `configure`). Tested with 4.08.1 and 4.09.0


samoht:
I haven't spotted any obvious issues, but it's hard to say with these scripts :-) So I'm good, as long as the CI are happy...

mato:
I'll take a detailed look next week.

We are missing CI for non-Linux platforms, so this is somewhat fragile.
FWIW, Cirrus CI supports at least FreeBSD, worth integrating here and
possibly elsewhere? Patches welcome :-)

https://cirrus-ci.org/guide/FreeBSD/

hannesm:
what I learned from doing these "cleanups" (motivated by e.g. #63) is that once we decide to drop `< 4.08.0` we can simplify ocaml-freestanding quite drastically (removing all the generated and checked in makefiles and m.YY.h -- the `s.h` still being provided by ocaml-freestanding and copied over to the OCaml directory). And that I'm not sure about the current 4.08.0+ support in this repository, as the OCaml build system changed quite radically (moved to autotools which implied renamings, it seems we're lucky and provide the CFLAGS via OC_CFLAGS) -- I'd be interested if there's any argument against using OCaml's `./configure` while building ocaml-freestanding.

with OCaml 4.10 being frozen, and MirageOS subscribed to "the last two major releases", I personally don't have any issue dropping all < 4.08.0 pretty soon (although we should "stay in sync" with xen for now to have some common supported runtimes).

mato:
@hannesm Mind re-basing against latest master to get your Cirrus CI for FreeBSD?

hannesm:
@mato thanks for your review. I can remove the calling `configure` commit from this PR, and have that as a separate PR?

I agree it is wrong to call configure this way (but also argue that the current checked-in configure output is brittle as seen in #63, and only works by accident).

mato:
> I can remove the calling configure commit from this PR, and have that as a separate PR?

Let's just do it here, I have something that looks like it may work.

mato:
@hannesm Take a look at cca57cb. Assuming CI passes, I think this is a workable solution for fully using autotools when on OCaml >= 4.08.0.

Makefile
```diff
@@ -36,32 +36,59 @@ build/ocaml/Makefile:
...
-endif
+	cd build/ocaml && \
+	    CC="cc $(OCAML_CFLAGS) -nostdlib" \
+	    AS="as" \
+	    ASPP="cc $(OCAML_CFLAGS) -c" \
+	    LD="ld" \
+	    CPPFLAGS="$(OCAML_CFLAGS)" \
+	    ./configure --host=x86_64-unknown-none \
+	        --disable-debugger --disable-systhreads --disable-unix-lib --without-libunwind
+	echo "ARCH=$(BUILD_ARCH)" >> build/ocaml/Makefile.config
```
hannesm: I suggest to retain the "freestanding SYSTEM", i.e. re-adding here:
` 	echo 'SYSTEM=freestanding' >> build/ocaml/Makefile.config`

mato: Why? There's nothing in the OCaml build system that refers to it, and the actual setting happens in https://github.com/ocaml/ocaml/blob/trunk/configure.ac#L850, which ends up with `none`. That block itself is problematic, since it's the reason we have to override `ARCH`, but that's a bug I should discuss with the OCaml compiler people.

Makefile
```diff
@@ -36,32 +36,59 @@ build/ocaml/Makefile:
...
-	cp config/m.$(BUILD_ARCH).h build/ocaml/byterun/caml/m.h
-	cp config/Makefile.$(BUILD_OS).$(BUILD_ARCH) build/ocaml/config/Makefile
-endif
+	cd build/ocaml && \
+	    CC="cc $(OCAML_CFLAGS) -nostdlib" \
+	    AS="as" \
+	    ASPP="cc $(OCAML_CFLAGS) -c" \
+	    LD="ld" \
+	    CPPFLAGS="$(OCAML_CFLAGS)" \
+	    ./configure --host=x86_64-unknown-none \
```
hannesm: for `--host=`, we should use `${BUILD_ARCH}-unknown-freestanding` (otherwise, arm64 won't get the right values). Using `freestanding` instead of `none` should lead to `Sys.os_type` returning `freestanding` again (as in earlier codebases). This implies that the `touch` below should be on `freestanding.c` instead of `none.c`.

I'll look more deeply and try this out later today. Thanks!

mato: It's complicated.

Yes, we should use the actual build (CPU) architecture in the `configure` invocation, thanks for checking that. However, it should probably be the architecture returned by `$(uname -m)` or out of `cc -dumpmachine`, not `BUILD_ARCH`, which is using architecture naming specific to OCaml (`arm64` for example).

This reminds me that Travis now has public support for ARM64, so I should add that to the CI mix here.

As for `freestanding` and the `configure` invocation:

1. We can't use `./configure --host=ARCH-unknown-freestanding` since this will confuse `configure`, GNU autotools terminology for "freestanding" is precisely `ARCH-unknown-none`.
2. I'm not sure what is asking for the dependency on `none.o` in `libasmrun.a`, will check. `OCAML_OS_TYPE` is just hard-coded to `Unix` via `ostype`, here: https://github.com/ocaml/ocaml/blob/trunk/configure.ac#L54.
3. As for what `Sys.os_type` returns, I think that is entirely arbitrary as long as it's not `Unix` or `Windows`. Code used by MirageOS applications should be using MirageOS APIs to determine this sort of thing, if at all.

hannesm:
I tested locally (thanks @mato) and have some observations and questions:
1) the resulting `libasmrun.a` is much bigger (1.5 MB instead of 470KB) -- since `-g` is now passed -- a `strip` on `libasmrun.a` pushes it back to ~200KB (this is only an observation, it is fine)
2) the arguments passed to configure -- I'm not entirely sure which ones are needed and which ones are not (we're only using the `libasmrun.a` target) -- for me `./configure --host=$(BUILD_ARCH)-unknown-none` (i.e. no further arguments) works as well and produces the same `libasmrun.a`
3) the amount of conditionals in Makefiles and shell scripts for 4.06 and 4.07 support let's me wonder whether we should just drop that -- users who require these compilers can use older ocaml-freestanding releases (e.g. 0.4.7).

Makefile
hannesm: ok, thanks for explanation, this is fine with me.

mato:
>     1. the resulting `libasmrun.a` is much bigger (1.5 MB instead of 470KB) -- since `-g` is now passed -- a `strip` on `libasmrun.a` pushes it back to ~200KB (this is only an observation, it is fine)

This is fine and expected (`configure` defaults to `-g`).

>     2. the arguments passed to configure -- I'm not entirely sure which ones are needed and which ones are not (we're only using the `libasmrun.a` target) -- for me `./configure --host=$(BUILD_ARCH)-unknown-none` (i.e. no further arguments) works as well and produces the same `libasmrun.a`

That was me experimenting with `configure` invocations. Removed.
 
>     3. the amount of conditionals in Makefiles and shell scripts for 4.06 and 4.07 support let's me wonder whether we should just drop that -- users who require these compilers can use older ocaml-freestanding releases (e.g. 0.4.7).

We can surely drop this, perhaps one release after the one with this change?

hannesm:
> We can surely drop this, perhaps one release after the one with this change?

fine with me. just keeping it in my head, before investing more time into the 4.06/4.07 compatibility, just remove it. since this PR works on 4.06/4.07, I see no reason to drop it immediately.

From my point of view & review, this PR is ready to be merged (if CI passes). if you like to squash all commits / reorder / squash some commits, please go ahead. Also feel free to merge at your convenience.

PR #66, opened 2019-11-22
Provide libdir, ld and ldflags into the distribution to be usable by dune

dinosaure:
Some information are missed (about #50). This PR add some others file to be usable by `dune` and compile/link correctly unikernels. Into details, `dune` with or without `duniverse` is able to load some information available from files.

In your case, `dune` can loads files located into `$(opam config var lib)/ocaml-freestanding/*` (with expansion of `%{lib:ocaml-freestanding:*}`). Currently, only `cflags` and `libs` are provided. This PR adds `ld`, `libdir` and `ldflags`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ flags/libs.tmp: flags/libs.tmp.in
 
 flags/libs.sexp: flags/libs.tmp Makeconf
 	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
-	    pkg-config $(PKG_CONFIG_DEPS) --libs >> $<
+	    pkg-config $(PKG_CONFIG_DEPS) --static --libs >> $<
 	awk -v RS= -- '{ \
 	    sub("@@PKG_CONFIG_EXTRA_LIBS@@", "$(PKG_CONFIG_EXTRA_LIBS)", $$0); \
 	    print "(", $$0, ")" \
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ FREESTANDING_LIBS=build/openlibm/libopenlibm.a \
 		  build/nolibc/libnolibc.a
 endif
 
-all:	$(FREESTANDING_LIBS) ocaml-freestanding.pc flags/libs flags/cflags
+all:	$(FREESTANDING_LIBS) ocaml-freestanding.pc flags/libs flags/cflags flags/ld flags/libdir flags/ldflags
 
 Makeconf:
 	./configure.sh
@@ -130,6 +130,22 @@ flags/cflags: flags/cflags.tmp Makeconf
 	awk -v RS= -- '{ \
 	    print "(", $$0, ")" \
 	    }' $< >$@
+
+flags/libdir: Makeconf
+	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
+		pkg-config $(PKG_CONFIG_DEPS) --variable=libdir >> $@
+	sed -i -e '1 s/$$/\/src/' $@
+
+flags/ld: Makeconf
+	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
+		pkg-config $(PKG_CONFIG_DEPS) --variable=libdir >> $@
+	sed -ni 's/\n/ /g' "$@"
+	if [ ! -s $@ ]; then echo "ld" >> $@; fi
+
+flags/ldflags: Makeconf
+	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
+		pkg-config $(PKG_CONFIG_DEPS) --variable=ldflags >> $@
+	sed -i "$@" -e 's/(\ )+/\n/g'
 
 install: all
 	./install.sh
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ flags/ld: Makeconf
 flags/ldflags: Makeconf
 	env PKG_CONFIG_PATH="$(shell opam config var prefix)/lib/pkgconfig" \
 		pkg-config $(PKG_CONFIG_DEPS) --variable=ldflags >> $@
-#	sed -i "$@" -e 's/(\ )+/\n/g'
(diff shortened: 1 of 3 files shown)


dinosaure:
'The patch differs from the initial patch of @TheLortex where a _sexp-expression_ is provided about `libdir` instead _raw arguments_. I did this choice to keep a consistency of provided files.

TheLortex:
Hi, thank you for this PR. I'm not sure using a _S-expression_ should work here.
Indeed there are several ways to import files in dune:
* `%{read:<file>}`, `%{read-lines:<file>}`,`%{read-strings:<file>}`: reads the file as a plain text file, this comes from variable expansion (https://dune.readthedocs.io/en/stable/concepts.html#variables)
* `(:include <file>)` parses a S-expression containing a single string (https://dune.readthedocs.io/en/stable/concepts.html#ordered-set-language).

So they are not part of the same language, and precisely I'm not sure that `(:include <file>)` can be used instead of `%{read:<file>}` for **rule actions**.

As a reminder, here are how there different files are used for the mirage dunification project:
* S-exp `cflags`: https://github.com/TheLortex/mirage/blob/full-dune/lib/mirage.ml#L744
* S-exp `libs`: https://github.com/TheLortex/mirage/blob/full-dune/lib/mirage.ml#L784
* plain `libdir`: https://github.com/TheLortex/mirage/blob/full-dune/lib/mirage.ml#L652
* plain `ld`/`ldflags`: https://github.com/TheLortex/mirage/blob/full-dune/lib/mirage.ml#L669 / https://github.com/TheLortex/mirage/blob/full-dune/lib/mirage.ml#L694

dinosaure:
> So they are not part of the same language, and precisely I'm not sure that (:include <file>) can be used instead of %{read:<file>} for rule actions.

My idea then, specially about the _dunification_ of MirageOS is the do the translation between arguments provided by these files to _s-expression_ by generated (by `mirage` tool) `dune` file. Again, I would like to keep a certain consistency between all files provided by `ocaml-freestanding` and avoid special layout needed only by `dune`.

An other solution is to explicitly provide cflags **.sexp** and libs **.sexp** to notice a special format of them by the extension.

hannesm:
About this PR and #50, they make me slightly uneasy due to duplication of information (in .pc and flags/). In a Mirage3 world (ocamlbuild), `pkg-config` and `.pc` files are used to pass around cflags/ldflags/.. . Now, in a Mirage4 world (dune), the same information is embedded in separate files, and pkg-config is still around, called at installation time of solo5 + ocaml-freestanding. Is this correct?

Would another path be implement the required pkg-config parts as OCaml library?
Will the result be `.pc` file free, and not depending on pkg-config (highly appreciated since we already discovered some divergence between Ubuntu, other Linux, and OpenBSD (where we install a custom symbolic link IIRC))?

dinosaure:
> About this PR and #50, they make me slightly uneasy due to duplication of information (in .pc and flags/). In a Mirage3 world (ocamlbuild), pkg-config and .pc files are used to pass around cflags/ldflags/.. . Now, in a Mirage4 world (dune), the same information is embedded in separate files, and pkg-config is still around, called at installation time of solo5 + ocaml-freestanding. Is this correct?

It's correct, but according generation of these files, we still follow what is provided by `ocaml-freestanding.pc`. I mean, with this PR, we have two paths for same values.

> Would another path be implement the required pkg-config parts as OCaml library?
Will the result be .pc file free, and not depending on pkg-config (highly appreciated since we already discovered some divergence between Ubuntu, other Linux, and OpenBSD (where we install a custom symbolic link IIRC))?

From what I know about libraries (like `bigstringaf` or `digestif`), they can get these files without `pkg-config` with, like I said, expansion of `%{lib:ocaml-freestanding:*}` - but the requirement is that these libraries must use `dune`.

However, in other side, I'm not sure that is easily possible to get off `pkg-config` from the `mirage` tool where some `*.pc` have the `Requires` field (which can not be imitated by the second way - the way implemented by this PR). Currently, I really would like to post-pone your concern after the _dunification_ and keep a _compatible_ eco-system first with these two ways.

Then, we can think and do the choice about the right direction (and surely properly delete `pkg-config`) with `dune` in our mind and our _unikernels_.

dinosaure:
Another diff, between files provided in this PR and `pkg-config` is that flags provided by files should be used by `ocamlopt` (with `-ccopt`/`-cclib`, see [this commit](https://github.com/mirage/ocaml-freestanding/pull/66/commits/aaba87a2dc68316c05d98c3365914ba187823da4)). And `pkg-config` provides flags for `ld`/`cc`.

About this specific question, I try to find an uniform way which concerns only (I think) `mirage` tool.

mato:
Stepping back a bit to get the big picture -- the goal here is to somehow integrate `dune`, `mirage` and _unikernels_. If we can replace `pkg-config` with something different, then we should try and do that.

As far as I can tell, the dependencies on `pkg-config` in the whole Mirage(/Solo5) stack are:

- `solo5-bindings-*`
    - `ocaml-freestanding`
        - `mirage-solo5`
            - (indirectly) the `mirage` tool and generated _unikernel_ build system
        - `gmp-freestanding`
        - `zarith-freestanding`
        - (unclear, indirect?) `nocrypto` / `mirage-entropy`

and we control all of these packages, we can change the whole stack as we see fit.

Now, I built/inherited the current `pkg-config` solution, so I know what pieces of information are passed around and how. What I don't know is:

**Q**: What are the _available mechanisms_ for a `dune`-based project/package _X_ to query dependency _Y_ for some key/value based metadata published by _X_, which is then "expanded" by `dune` during build of package _X_?

In other words, are we limited to expansion of the content of "some file published by X"? How does the location of "some file" get determined? What formats of "some file" can `dune` parse? Can this mechanism also work in some sensible way _without_ depending on `dune`? For example, "cross"-builds of C packages such as `gmp-freestanding`?

Without understanding what's available I can't offer any potential solutions / replacements for `pkg-config`.

dinosaure:
I did mirage/mirage#1018 to not lose what we talk in this issue. Please continue to this issue instead this PR.

mato:
@dinosaure As I understand it, this PR is currently out of date / "to be replaced by something else"? Or should it be considered for merging?

dinosaure:
Out of date 👍 !

PR #67, opened 2019-11-22
add cirrus with two FreeBSD runners: one on 4.08.1, one on 4.09.0

hannesm:
they both use solo5-bindings-hvt


mato:
Thanks, this is good!

Issue #71, opened 2020-02-11
could endian.h be enhanced with htobe64 and be64toh


hannesm:
these primitives are needed for mirage-crypto

(e.g. from https://github.com/dune-universe/ocaml-nocrypto/commit/3d7bdd58fe8b46d7d14732e8dbe6e8a0ede0d306):
```C
#if (BYTE_ORDER == LITTLE_ENDIAN)
#define htobe64(x) (_bswap64(x))
#define be64toh(x) (_bswap64(x))
#elif (BYTE_ORDER == BIG_ENDIAN)
#define htobe64(x) (x)
#define be64toh(x) (x)
#else
#error "Cannot determine byte order"
#endif
```

hannesm:
further investigation shows that on FreeBSD and OpenBSD the solo5 configure script copies sys/endian.h (and friends, machine/endian.h, x86/endian.h) into the include directory already (which may have an impact on this request).

hannesm:
and this is not very high priority - there's an implementation already in a separate file of mirage-crypto. though, I think it would be sensible to have the "full" endian.h supported by freestanding, especially the little-and-big-endian-to-host conversions, plus defining BYTE_ORDER:
```C
htobe16
htobe32
htobe64
htole16
htole32
htole64

be16toh
be32toh
be64toh
le16toh
le32toh
le64toh
```

mato:
The way to do this would be to port the relevant bits from musl, since that is what nolibc is primarily based on. Given that you have mentioned here and elsewhere that this is not a priority for you / you have other ways of accomplishing the same result, I'm not going to look at this for the next release.

hannesm:
I'll close this issue, since I don't need these functions.

PR #73, opened 2020-04-27
Provide byte-swapping functions in endian.h, pass -D__ocaml_freestanding__ via CFLAGS

hannesm:
the code of endian.h is taken from musl, and adapted to make clang's warnings go away.

in addition, `-D__freestanding__` is passed via CFLAGS, which allows other C code to use `#if defined(__freestanding__)` and conditionally do something.

An example use of both endian.h and the __freestanding__ is https://github.com/hannesm/mirage-crypto/commit/a66190dccf9c596c24dfc4a45a911ef9effafab7 (which compiles fine with this change and the corresponding solo5 PR).

I'm interested in your opinion, and am open for suggestions: given that BYTE_ORDER will always be LITTLE_ENDIAN, is it worth to keep the 12 lines of defines? should `-D __freestanding__` be renamed to something else?

This PR allows us to unify the copied headers of solo5 a bit more (see https://github.com/Solo5/solo5/pull/453), and has been requested in #71. It will as well help us to eventually resolve https://github.com/mirage/hacl/issues/30.

### DIFF
--- a/flags/cflags.tmp.in
+++ b/flags/cflags.tmp.in
@@ -1,1 +1,1 @@
--D__freestanding__ -I%{prefix}%/include/ocaml-freestanding
+-D__ocaml_freestanding__ -I%{prefix}%/include/ocaml-freestanding
--- a/flags/cflags.tmp.in
+++ b/flags/cflags.tmp.in
@@ -1,1 +1,1 @@
--D__ocaml_freestanding__ -I%{prefix}%/include/ocaml-freestanding
+-U__linux__ -U__linux -Ulinux -U__gnu_linux__ -U__FreeBSD__ -U__OpenBSD__ -D__ocaml_freestanding__ -I%{prefix}%/include/ocaml-freestanding
--- a/flags/cflags.tmp.in
+++ b/flags/cflags.tmp.in
@@ -1,1 +1,1 @@
--I%{prefix}%/include/ocaml-freestanding
+-D__freestanding__ -I%{prefix}%/include/ocaml-freestanding
--- a/nolibc/include/endian.h
+++ b/nolibc/include/endian.h
@@ -9,4 +9,57 @@
 #error Unsupported architecture
 #endif
 
+static __inline uint16_t __bswap16(uint16_t __x)
+{
+  return __x<<8 | __x>>8;
+}
+
+static __inline uint32_t __bswap32(uint32_t __x)
+{
+  return __x>>24 | (__x>>8&0xff00) | (__x<<8&0xff0000) | __x<<24;
+}
+
+static __inline uint64_t __bswap64(uint64_t __x)
+{
+  return (__bswap32(__x)+0ULL)<<32 | __bswap32(__x>>32);
+}
+
+#define bswap16(x) __bswap16(x)
+#define bswap32(x) __bswap32(x)
+#define bswap64(x) __bswap64(x)
+
+#if __BYTE_ORDER == __LITTLE_ENDIAN
+
+#define htobe16(x) __bswap16(x)
+#define htobe32(x) __bswap32(x)
+#define htobe64(x) __bswap64(x)
+#define htole16(x) (uint16_t)(x)
+#define htole32(x) (uint32_t)(x)
+#define htole64(x) (uint64_t)(x)
+
+#define be16toh(x) __bswap16(x)
+#define be32toh(x) __bswap32(x)
+#define be64toh(x) __bswap64(x)
+#define le16toh(x) (uint16_t)(x)
+#define le32toh(x) (uint32_t)(x)
+#define le64toh(x) (uint64_t)(x)
+
+#else /* BYTE_ORDER != __LITTLE_ENDIAN */
+
+#define htobe16(x) (uint16_t)(x)
+#define htobe32(x) (uint32_t)(x)
+#define htobe64(x) (uint64_t)(x)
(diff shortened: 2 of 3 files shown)


mato:
> I'm interested in your opinion, and am open for suggestions: given that BYTE_ORDER will always be LITTLE_ENDIAN, is it worth to keep the 12 lines of defines?

I'd keep them. In fact, the compiler predefines a `__BYTE_ORDER__` (see https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html, also present on clang), so the arch-dependent block of #ifs can be modified suitably to depend only on the compiler's idea of endianness, which makes endian.h entirely generic and future-proof, which is good.

> should -D __freestanding__ be renamed to something else?

Dunno. __freestanding__ seems fine to me.

hannesm:
> >  should `-D __freestanding__` be renamed to something else?

> Dunno. `-D__freestanding__` seems fine to me.

After thinking about it more, I renamed it to `-D __mirage_freestanding__`. For people from the outside (who are not familiar with MirageOS internals), it is easier to relate to something where `mirage` is present in the name.

mato:
> After thinking about it more, I renamed it to -D__mirage_freestanding__. For people from the outside (who are not familiar with MirageOS internals), it is easier to relate to something where mirage is present in the name.

If it's not just `__freestanding__` then if you don't mind I'd prefer `__ocaml_freestanding__` since that matches the name of the package that introduces the flag. Googling for "ocaml freestanding" leads here, and by extension to Mirage, so people will make the association anyway.

hannesm:
@mato sure, done

mato:
@hannesm 

I wonder if there's a way to make this work both in the presence and _absence_ of Solo5/solo5#453 -- the problem I see is that it's technically an API-breaking change between layers, but I'd like to avoid going through the whole opam upper/lower bounds dance for it.

I started writing up notes on the various possible failure cases, but then got distracted by a video call and have lost most of the context now. One case I can remember is:

If there is some downstream consumer that, on a FreeBSD build host with a released version of Solo5 and ocaml-freestanding with this change, includes (transitively or otherwise) both `sys/endian.h` (getting the system one mistakenly provided by Solo5) and `endian.h` (getting this new one), then the compilation will fail due to various duplicate definitions.

What would help a lot is if you could try and track down all the _currently released_ consumers of a `endian.h` _or_ `sys/endian.h` downstream of ocaml-freestanding. i.e. is it potentially used by existing releases of `gmp-freestanding`? `mirage-cypto`? `ocaml-nocrypto`?

That would at least help identify if we need to do something about the failure cases/API break or if we can mostly ignore it.

hannesm:
I could find the following users of endian.h - nocrypto, mirage-crypto, digestif. They are all using the same copied-around [`bitfn.h`](https://github.com/mirage/mirage-crypto/blob/v0.6.2/src/native/bitfn.h#L161-L180) which has an `#ifdef` chain based on `__linux__` / `__FreeBSD__` -- without a case for `__ocaml_freestanding__` --> the current released packages will continue with a release of ocaml-freestanding with this PR merged.

They will all fail with an updated solo5 (on BSD systems, due to lack of `sys/endian.h`). The only way to mitigate this as far as I can tell is to add `conflict` statements to the existing packages for `"solo5-*-hvt" {> "0.6.4"} (os = "freebsd" | os = "openbsd")`.

Safety of this PR: it may lead to multiple defined symbols _if and only if_ something `#include <endian.h>` and `#include <sys/endian.h>`. I have not observed any library doing this in the wild. To mitigate such a situation, I would be fine to include:
```C
#ifdef _SYS_ENDIAN_H //both FreeBSD and OpenBSD sys/endian.h define this.
#error "<sys/endian.h> already included while attempting to include ocaml-freestanding <endian.h>"
#else
... our endian.h ...
#endif
```

Once this PR is merged and released, there need to be update to digestif and mirage-crypto (to continue to work on BSD systems - which will then depend on ocam-freestanding > 0.5.0), the smallest possible patch is in bitfn to start with:
```C
#ifdef __ocaml_freestanding__
#include <endian.h>
... reset of elif chain ...
```

hannesm:
I added a commit which undefines the definitions mentioned above. The advantage of doing so is many-fold: the compilation of C sources with ocaml-freestanding is more uniform across platforms; and the breakage (in respect to nocrypto/mirage-crypto/digestif -- bitfn.h) is non-existing. In addition, the hacl package (which was the motivation to look into this issue for me) will just work as is (since their [`#ifdef` chain](https://github.com/mirage/hacl/blob/v0.1.0/src/kremlib.h#L191-L330) respects the C compiler `__BYTE_ORDER__` as a last resort (which is now hit with ocaml-freestanding on any platform)). In mirage-crypto I removed all the `#ifdef` in https://github.com/mirage/mirage-crypto/pull/56 and now depend on `__BYTE_ORDER__` defined by the C compiler.

It is also more the truth of ocaml-freestanding -- this is neither linux nor FreeBSD nor anything else (and C code that uses `#ifdef` chains to select the platform should better fail hard (at compilation) when freestanding is encountered than soft (at link time).

Given the current mess of POSIX and endian.h (where to put it, which symbols to export), maybe we should just not provide it in freestanding at all?

mato:
@hannesm Thank you very much for looking into this, I appreciate it. I've done a V2 inspired by your changes, but with a few other improvements in #74. Continuing the discussion there.

hannesm:
superseeded by #74

PR #74, opened 2020-05-04
V2: Define __ocaml_freestanding__, fix the endian.h mess

mato:
This is a re-work of #73, containing the following changes:

* nolibc: Provide a sys/param.h with BYTE_ORDER (~~83fc2b4~~ 3226548)
* nolibc: Provide a full endian.h (~~3c251fe~~ 5055052)
* nolibc: Define `__ocaml_freestanding__`, add overrides.h (239263b)

See individual commits for details.

/cc mirage/hacl#30 mirage/mirage-crypto#54 Solo5/solo5#453 @hannesm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ Makeconf:
 	./configure.sh
 
 TOP=$(abspath .)
-FREESTANDING_CFLAGS+=-isystem $(TOP)/nolibc/include
+FREESTANDING_CFLAGS+=-isystem $(TOP)/nolibc/include -include _freestanding/overrides.h
 
 build/openlibm/Makefile:
 	mkdir -p build/openlibm
--- a/flags/cflags.tmp.in
+++ b/flags/cflags.tmp.in
@@ -1,1 +1,1 @@
--I%{prefix}%/include/ocaml-freestanding
+-I%{prefix}%/include/ocaml-freestanding -include _freestanding/overrides.h
--- a/nolibc/include/_freestanding/byteorder.h
+++ b/nolibc/include/_freestanding/byteorder.h
@@ -1,0 +1,16 @@
+#ifndef __FREESTANDING_BYTEORDER_H
+#define __FREESTANDING_BYTEORDER_H
+
+#if !defined(__BYTE_ORDER__)
+#error C compiler does not define __BYTE_ORDER__
+#endif
+
+#if defined(BYTE_ORDER) || defined(LITTLE_ENDIAN) || defined(BIG_ENDIAN)
+#error BYTE_ORDER, LITTLE_ENDIAN or BIG_ENDIAN must not be defined here
+#endif
+
+#define BYTE_ORDER __BYTE_ORDER__
+#define LITTLE_ENDIAN __ORDER_LITTLE_ENDIAN__
+#define BIG_ENDIAN __ORDER_BIG_ENDIAN__
+
+#endif /* __FREESTANDING_BYTEORDER_H */
--- a/nolibc/include/_freestanding/overrides.h
+++ b/nolibc/include/_freestanding/overrides.h
@@ -1,0 +1,13 @@
+#ifndef __FREESTANDING_OVERRIDES_H
+#define __FREESTANDING_OVERRIDES_H
+
+#undef __FreeBSD__
+#undef __OpenBSD__
+#undef __gnu_linux__
+#undef __linux
+#undef __linux__
+#undef linux
+
+#define __ocaml_freestanding__
+
+#endif /* __FREESTANDING_OVERRIDES_H */
--- a/nolibc/include/endian.h
+++ b/nolibc/include/endian.h
@@ -1,12 +1,45 @@
 #ifndef _ENDIAN_H
 #define _ENDIAN_H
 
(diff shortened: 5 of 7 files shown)


mato:
@hannesm:

I have tested this manually on Debian 10 with GCC, including inspecting the `config.log` from the OCaml runtime and libgmp (gmp-freestanding) to verify that they are doing the right thing. Note that neither of these use an `<endian.h>` or `<sys/endian.h>`.

Regarding your comments in #73:

> It is also more the truth of ocaml-freestanding -- this is neither linux nor FreeBSD nor anything else (and C code that uses #ifdef chains to select the platform should better fail hard (at compilation) when freestanding is encountered than soft (at link time).

Agreed. This V2 does that via `-include _freestanding/overrides.h`, IMO cleaner than adding the various undefs to CFLAGS in multiple places. It also applies the same flags to code built as part of ocaml-freestanding itself, notably the OCaml runtime.

> Given the current mess of POSIX and endian.h (where to put it, which symbols to export), maybe we should just not provide it in freestanding at all?

We may as well provide an `<endian.h>` and relevant [interfaces](https://www.austingroupbugs.net/view.php?id=162). That way, downstreams that want portable code with minimal duplication can do:

    #if defined(__ocaml_freestanding__)
    #include <endian.h>
    #elif defined(...)
    /* ... */
    #endif     

... I see no harm in that.

The remaining question for me is that from your comments on #73 it's no longer obvious to me if we need the `#ifndef _SYS_ENDIAN_H` hack in our `endian.h` at all? 

If you look at `nolibc/include/_freestanding/byteorder.h`, I've arranged things so that _were we to remove the `_SYS_ENDIAN_H` guard_, then in the presence of that particular failure case compilation will fail hard due to `BYTE_ORDER` already being defined. My feeling is it's better to deal with (and report/fix) breakage rather than paper over the problem, but I'm not sure.

What do you think?

hannesm:
This looks fine to me, I tested it on a FreeBSD 12.1 system.

> The remaining question for me is that from your comments on #73 it's no longer obvious to me if we need the `#ifndef _SYS_ENDIAN_H` hack in our `endian.h` at all?

I don't think we need it. The undefine of the host platform is sufficient to avoid issues in the common path, and there's no need to support "broken" `#ifdef .. #elif .. #endif` expressions.

After this is merged, I only have two more questions (out of scope for this PR I suspect, but since we're cleaning up headers, I though I'd raise them here):
- why do we copy only a selected set of `.h` files from the OCaml runtime? I'd expect a `cp -R runtime/caml/ ${DESTINC}/caml/` to be a much more sustainable solution (see #75 for what I mean)
- the `openlibm` header file issue - our install.sh copies openlibm/include and openlibm/src/*h (this is the same as openlibm's `make install-headers`) into `${DESTINC}` which pollutes this quite a bit with private headers. My attempts to instead install into `${DESTINC}/openlibm` were not too successful (the zarith-freestanding fails since it can't find the openlibm-dependent headers), but I'd appreciate a solution where our `math.h` symlinks to `openlibm/openlibm.h` and everything just works (i.e. not using `-I${DESTINC}/openlibm`).

mato:
> This looks fine to me, I tested it on a FreeBSD 12.1 system.

Thanks

> I don't think we need it. The undefine of the host platform is sufficient to avoid issues in the common path, and there's no need to support "broken" `#ifdef .. #elif .. #endif` expressions.
> 
> EDIT: I pushed [8c06c0f](https://github.com/mirage/ocaml-freestanding/commit/8c06c0faad0b71e758f5cc90daaca8297d4d0df1) to your branch that removes the `_SYS_ENDIAN_H` guard / definition.

Ok. I rebased the existing commits, removing the `_SYS_ENDIAN_H` business entirely. No need to keep the back-and-forth in the history.

> After this is merged, I only have two more questions (out of scope for this PR I suspect, but since we're cleaning up headers, I though I'd raise them here):
> 
>     * why do we copy only a selected set of `.h` files from the OCaml runtime? I'd expect a `cp -R runtime/caml/ ${DESTINC}/caml/` to be a much more sustainable solution (see #75 for what I mean)

No idea. I'm sure there was some reason for why it got done that way. It's ~300kB of headers and "stuff", if you're fine with just installing all of it, then let's do it.

>     * the `openlibm` header file issue - our install.sh copies openlibm/include and openlibm/src/*h (this is the same as openlibm's `make install-headers`) into `${DESTINC}` which pollutes this quite a bit with private headers. My attempts to instead install into `${DESTINC}/openlibm` were not too successful (the zarith-freestanding fails since it can't find the openlibm-dependent headers), but I'd appreciate a solution where our `math.h` symlinks to `openlibm/openlibm.h` and everything just works (i.e. not using `-I${DESTINC}/openlibm`).

Yeah, this doesn't work because the openlibm public headers want to include the private headers, and do so without using `#include "openlibm/...". I checked openlibm master and it's still that way, not sure what to do about it other than various `sed` hackery.

hannesm:
@mato great, let's just defer that openlibm to a later point in time (don't fix issues that are not present).

I looked into the history of the caml-header list, and it came with the initial commit. I feel more comfortable to install all of them (they are in `caml/` anyways, and don't conflict with other headers).

mato:
All green, good to merge!

PR #75, opened 2020-05-05
install all caml headers

hannesm:
instead of using a hardcoded list of selected headers


mato:
As discussed in #74, this is fine. Ignoring the flaky Arm64 builder, merging.

PR #76, opened 2020-05-06
Build system cleanups, provide minimal <inttypes.h>

### LINKED ISSUE #60: How to add support for <inttypes.h>
Hello,

I'm trying to build a library that uses `<inttypes.h>`, more specifically it's in [`hacl`](https://github.com/mirage/hacl/blob/54edddf123bd79861a9c76b053ddd6739ba9cc69/src/kremlib_base.h#L26) - the `PRIu64` macro [is used](https://github.com/mirage/hacl/blob/54edddf123bd79861a9c76b053ddd6739ba9cc69/src/kremlib.h#L446) - it should expand to a format string specifier. This C99 header are unfortunately not supported in `ocaml-freestanding`.

In this particular case I might be able to patch my way around this problem (this debug code can be stripped), but this is not ideal as this code is extracted, so I'd rather not touch it too much.

How would I go about adding support for `<inttypes.h>`? can I just add `inttypes.h` in `nolibc/include/`?

(same would apply to `bswap64` - I recall adding it to a couple libraries because it was missing in freestanding)

Thanks!

mato:
The primary goal here is to provide a minimal <inttypes.h>, which fixes #60. The associated changes improve the build system, notably adding compile-testing of all public headers provided by nolibc.

See individual commits for details:

- Build system cleanup, remove build/ (a5b14dd)
- nolibc: Compile-test public headers at build time (7ca57b9)
- nolibc: Provide a minimal <inttypes.h> (38a35e1)

/cc @hannesm @emillon @dinosaure (you asked for/could use an inttypes.h) @samoht (build system changes)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@ flags/libs
 flags/libs.tmp
 nolibc/*.o
 nolibc/*.a
+nolibc/test-include/
 ocaml/
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 Makeconf
-build/
-config/
 ocaml-freestanding.pc
 flags/cflags
 flags/cflags.tmp
 flags/libs
 flags/libs.tmp
+nolibc/*.o
+nolibc/*.a
+ocaml/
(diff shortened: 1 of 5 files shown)


mato:
Acked by @hannesm in #77, merging.

PR #79, opened 2020-06-11
Get rid of calls to opam when using opam

kit-ty-kate:


kit-ty-kate:
Sorry for the lack of description on this PR. The previous version of ocaml-freestanding is already installable with OCaml 4.11 (see https://github.com/ocaml/opam-repository/pull/16976). The only thing that this PR does is to make the build the tinyest bit faster and generally easier to install without opam.

OPAM 2.1 does not have to do anything wrt that, it was just a bug that has been fixed a while ago, but in my opinion it highlighted parts of packages that could be improved by making them the tiniest bit faster and more portable.

mato:
Ok, I get what this is supposed to do, but am going to close it for now; it introduces dependencies on a $PREFIX in the environment in ways that are not quite obvious, and in any case `opam config var prefix` takes about 0.013s to execute on my system.

PR #82, opened 2020-07-22
nolibc: Fix posix_memalign() return values

mato:
In the failure cases, posix_memalign() should return -1 and set errno
rather than returning an errno value.


mato:
Ignoring the FreeBSD 4.09.0 failure, it's a transient error.

PR #83, opened 2020-08-26
Travis, Cirrus CI: Update for OCaml 4.11

mato:
For Travis, test common archs and targets for the latest OCaml release, and only amd64/hvt for previous releases.

/cc @kit-ty-kate

PR #89, opened 2020-11-09
cirrus: update tested compiler versions

hannesm:
Dor each supported compiler, test the most recent release on FreeBSD. Adds
testing with OCaml 4.11.1.

This was partially part of #83, which fails due to Travis + OCaml CI scripts still not supporting OCaml 4.11. See https://discuss.ocaml.org/t/removing-fat-compiler-images-on-hub-docker-com-r-ocaml-opam2/ and https://github.com/ocaml/ocaml-ci-scripts/pull/347 for further information and potential workarounds.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,8 +7,9 @@ freebsd_task:
   env:
     matrix:
       - OCAML_VERSION: 4.08.1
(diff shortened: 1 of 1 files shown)


mato:
Thanks!

PR #91, opened 2020-12-15
Remove cflags & libs flags, installing pkg-config data as s-expressions

hannesm:
These files were added in #50 (v0.4.3) when we planned to use this path to
compile MirageOS. Time has changed, and there is a different plan now, which
makes these superfluous.

In the light of #90 this here is a self-contained small PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,5 @@
 Makeconf
 ocaml-freestanding.pc
-flags/cflags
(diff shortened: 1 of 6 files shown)


hannesm:
since #92 includes this commit, and it simplifies the build system - and tests are fine, I'll merge it.

PR #93, opened 2021-03-18
Make build system more flexible to aid with NixOS packaging

sternenseemann:
This PR contains two patches for the build system I wrote while giving packaging `ocaml-freestanding` for NixOS a shot. I wanted to make the following things possible:

* Building without the `ocaml-src` package by alternatively extracting the ocaml source archive to `./ocaml` manually
* Building without `opam`
* Overriding the auto detection mechanism which selects the solo5 binding to use

The commit messages have more details on the reasoning.

I tried to preserve the old default behavior, but can't say for sure I haven't broken anything. Someone with a more “normal” setup testing this would be appreciated!

I don't need a merge desparately, I can also vendor these patches in `nixpkgs`. If you are switching to `dune` anyways, it not be worth it to mess with the build system.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -11,13 +11,19 @@ if pkg_exists solo5-bindings-hvt solo5-bindings-spt solo5-bindings-virtio solo5-
     echo "ERROR: Only one of solo5-bindings-hvt, solo5-bindings-spt, solo5-bindings-virtio, solo5-bindings-muen, solo5-bindings-genode, solo5-bindings-xen can be installed." 1>&2
     exit 1
 fi
-PKG_CONFIG_DEPS=
-pkg_exists solo5-bindings-hvt && PKG_CONFIG_DEPS=solo5-bindings-hvt
-pkg_exists solo5-bindings-spt && PKG_CONFIG_DEPS=solo5-bindings-spt
-pkg_exists solo5-bindings-muen && PKG_CONFIG_DEPS=solo5-bindings-muen
-pkg_exists solo5-bindings-virtio && PKG_CONFIG_DEPS=solo5-bindings-virtio
-pkg_exists solo5-bindings-genode && PKG_CONFIG_DEPS=solo5-bindings-genode
-pkg_exists solo5-bindings-xen && PKG_CONFIG_DEPS=solo5-bindings-xen
+if [ -z "${PKG_CONFIG_DEPS}" ]; then
+    PKG_CONFIG_DEPS=
+    pkg_exists solo5-bindings-hvt && PKG_CONFIG_DEPS=solo5-bindings-hvt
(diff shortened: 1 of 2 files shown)


sternenseemann:
Seems like the build failure is the same as on master.

dinosaure:
I will try to see if we can integrate such great work on #92 which will be merged soon.

sternenseemann:
Seems like a good idea, I'll try to build that branch and see if there are any issues or what needs to be fixed.

I can imagine a lot would work out of the box already since we actually have dune available at build time and it is generally well-integrated. Packaging `ocaml-src` is also I think a hit worth taking if it allows for a cleaner dune package.

configure.sh
```diff
@@ -9,19 +11,25 @@ if pkg_exists solo5-bindings-hvt solo5-bindings-spt solo5-bindings-virtio solo5-
...
+        || (echo "ERROR: ${PKG_CONFIG_DEPS} is not installed" 1>&2; exit 1) \
+        || exit 1
+fi
 if [ -z "${PKG_CONFIG_DEPS}" ]; then
     echo "ERROR: No supported Solo5 bindings package found." 1>&2
     echo "ERROR: solo5-bindings-hvt, solo5-bindings-spt, solo5-bindings-virtio, solo5-bindings-muen, solo5-bindings-genode or solo5-bindings-xen must be installed." 1>&2
     exit 1
 fi
-ocamlfind query ocaml-src >/dev/null || exit 1
+[ -e "$(dirname "$0")/ocaml" ] || ocamlfind query ocaml-src >/dev/null || exit 1
```
dinosaure, line 32: It seems the only interesting part which can be add into #92. Others stuffs are deleted (due to the deletion of `pkg-config`).

sternenseemann: I think this check doesn't need to exist in the configure file at all actually since `ocaml-src` isn't used for anything here and the build will fail pretty early if it is missing.

The avantage of having it only in the Makefile potentially could be that make doesn't try to build it if it is already present.

hannesm:
Thanks for your work. Since this adds quite a lot of complexity (conditionals) to the build system, and additionally the "use dune to compile MirageOS" is on the horizon, I will not merge this, but instead hope that #92 will integrate the required changes (as @dinosaure pointed out, only small changes are needed).

dinosaure:
I will keep this PR opens even if we merged #104 to keep in our mind the NixOS support.

hannesm:
IMHO this can be closed -- in `configure.sh` ocamlfind is not used anymore, and in `Makefile` only if `ocaml/Makefile` is not present, it is copied (from `` `opam var prefix`/lib/ocaml-src ``). Please report if packaging for NixOS needs further changes in a separate issue (or PR). Thanks a lot for your patience.

PR #94, opened 2021-09-14
4.13 support

dra27:
A change made to OCaml before 4.13 was branched slowed down the Windows build, but the fix (although ready for review) was deferred until during the beta cycle. The story of the fix is long and not entirely relevant, but the solution is that there is now a one-file C program in `runtime/` which forms part of the _build system_ and so has to be built with a host ("build") C compiler. There should be an elegant solution to this in future when the OCaml tree itself natively supports cross-compilation, but in the meantime we're proposing inserting three variable hooks which ocaml-freestanding (and other cross-compiling setups) can use.

This can be tested with

```
opam pin add ocaml-src git+https://github.com/dra27/ocaml.git#freestanding-sak-4.13 -yn
opam pin add ocaml-freestanding git+https://github.com/dra27/ocaml-freestanding.git#4.13-sak -y
```

cc @kit-ty-kate


hannesm:
thanks, superseeded by #95

PR #95, opened 2021-09-24
OCaml 4.13 support

kit-ty-kate:
Simply #94 with the addition of CI to show it works.

OCaml 4.13.0 got released a few hours ago.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,9 @@ ocaml/Makefile.config: ocaml/Makefile
 	    CPPFLAGS="$(OCAML_CFLAGS)" \
 	    ./configure --host=$(BUILD_ARCH)-unknown-none
 	echo "ARCH=$(OCAML_BUILD_ARCH)" >> ocaml/Makefile.config
+	echo 'SAK_CC=cc' >> ocaml/Makefile.config
+	echo 'SAK_CFLAGS=$(OC_CFLAGS) $(OC_CPPFLAGS)' >> ocaml/Makefile.config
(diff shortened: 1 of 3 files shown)


hannesm:
thanks.

Makefile
```diff
@@ -46,6 +46,9 @@ ocaml/Makefile.config: ocaml/Makefile
 	    CPPFLAGS="$(OCAML_CFLAGS)" \
 	    ./configure --host=$(BUILD_ARCH)-unknown-none
 	echo "ARCH=$(OCAML_BUILD_ARCH)" >> ocaml/Makefile.config
+	echo 'SAK_CC=cc' >> ocaml/Makefile.config
+	echo 'SAK_CFLAGS=$(OC_CFLAGS) $(OC_CPPFLAGS)' >> ocaml/Makefile.config
+	echo 'SAK_LINK=$(SAK_CC) $(SAK_CFLAGS) $(OUTPUTEXE)$(1) $(2)' >> ocaml/Makefile.config
```
dinosaure, line 51: What is the purpose of these options?

hannesm: Thanks for your comment. The build system of OCaml changed due to `iconv` not available on Windows (see https://github.com/ocaml/ocaml/pull/10451). Now, it ships and uses during build a small C utility (called `sak`) that performs the job `iconv` used to do. Thanks to last-minute addition in https://github.com/ocaml/ocaml/pull/10635 we can use the host C compiler only for `sak` (which we do here) to succeed the build (since the modified C compiler & flags of our cross-build is not able to compile, link, and execute `sak`).

I hope this answers your question.

dinosaure: Thanks, I see, I re-updated #92 with the current status of `master` but I would like a double-check from you and @TheLortex.

Issue #101, opened 2021-12-17
ocaml-freestading 0.6.6 failing to compile on OpenBSD (7.0 Current)


adamsteen:
```
checking for copysign... yes
configure: error: C99 float ops unavailable, enable replacements with --enable-imprecise-c99-float-ops
gmake: *** [Makefile:43: ocaml/Makefile.config] Error 1
```

```
$ ocaml --version
The OCaml toplevel, version 4.13.1
$ opam --version
2.0.8
```

if there is anything else you need or i can help with please let me know

dinosaure:
That weird because I believed that this PR (included into 6.6) fixed the error: https://github.com/mirage/ocaml-freestanding/pull/99 can you check with `~master`? May be we did something wrong with the release.

adamsteen:
@dinosaure whats the best way to pin `~master`, sorry i am more a c developer?

adamsteen:
```
$ opam pin add git+https://github.com/mirage/ocaml-freestanding.git
[ocaml-freestanding] synchronised from git+https://github.com/mirage/ocaml-freestanding.git
[ocaml-freestanding.0.7.0] synchronised from git+https://github.com/mirage/ocaml-freestanding.git
ocaml-freestanding is now pinned to git+https://github.com/mirage/ocaml-freestanding.git (version 0.7.0)
The following actions will be performed:
  - install   ocaml-freestanding 0.7.0*
  - recompile io-page            2.4.0  [uses ocaml-freestanding]
  - recompile mirage-unix        4.0.1  [uses io-page]
===== 1 to install | 2 to recompile =====
Do you want to continue? [Y/n] y

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[io-page.2.4.0] found in cache
[mirage-unix.4.0.1] found in cache
Processing  3/3: [ocaml-freestanding.0.7.0: git]
[ocaml-freestanding.0.7.0] synchronised from git+https://github.com/mirage/ocaml-freestanding.git

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The compilation of ocaml-freestanding failed at "/usr/local/bin/gmake".

#=== ERROR while compiling ocaml-freestanding.0.7.0 ===========================#
[...]

PR #102, opened 2021-12-30
Use OCaml's configure with LDFLAGS and libs from solo5 & freestanding

hannesm:
See individual commits for details, this exposes #101 on all platforms.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,12 @@ FREESTANDING_LDFLAGS+=
 openlibm/libopenlibm.a:
 	$(MAKE) -C openlibm "CFLAGS=$(FREESTANDING_CFLAGS)" libopenlibm.a
 
+# OCaml configure injects -lm into LIBS if it can be linked. This leads to
+# inconsistent assumptions - the host system libm being used and openlibm.
+# Since this is hardcoded in configure, we modify configure to never add -lm.
 ocaml/Makefile:
 	cp -r `ocamlfind query ocaml-src` ./ocaml
+	sed -i -e 's/ac_cv_lib_m_cos=yes/ac_cv_lib_m_cos=no/' ocaml/configure
 
 # OCaml >= 4.08.0 uses an autotools-based build system. In this case we
 # convince it to think it's using the Solo5 compiler as a cross compiler, and
--- a/Makefile
+++ b/Makefile
@@ -37,13 +37,13 @@ ocaml/Makefile:
 #   link against nolibc (which would need to produce complete Solo5 binaries).
 # - We override OCAML_OS_TYPE since configure just hardcodes it to "Unix".
 OCAML_CFLAGS=$(FREESTANDING_CFLAGS) -I$(TOP)/openlibm/include -I$(TOP)/openlibm/src
-OCAML_LDFLAGS=$(FREESTANDING_LDFLAGS) -L$(TOP)/openlibm/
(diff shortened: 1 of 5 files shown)


hannesm:
The last force-push now patches OCaml's configure to avoid passing `-lm` for conftest. Instead, the `-lopenlibm` is used. This allows configure to execute nicely.

Makefile
```diff
@@ -53,9 +52,6 @@ ocaml/Makefile.config: ocaml/Makefile
 	echo 'SAK_CC=cc' >> ocaml/Makefile.config
 	echo 'SAK_CFLAGS=$(OC_CFLAGS) $(OC_CPPFLAGS)' >> ocaml/Makefile.config
 	echo 'SAK_LINK=$(SAK_CC) $(SAK_CFLAGS) $(OUTPUTEXE)$(1) $(2)' >> ocaml/Makefile.config
-	echo '#define HAS_GETTIMEOFDAY' >> ocaml/runtime/caml/s.h
-	echo '#define HAS_SECURE_GETENV' >> ocaml/runtime/caml/s.h
-	echo '#define HAS_TIMES' >> ocaml/runtime/caml/s.h
```
dinosaure, line 58: Why we need to remove these lines?

dinosaure: ah ok I saw the commit. I will do the end to end test on Ubuntu and just a compilation on OpenBSD 7.

hannesm: since OCaml's configure is now using nolibc and openlibm, it properly detects these features now. tested on 4.13.0.

dinosaure:
It seems that this version of `ocaml-freestanding` brings some _syscalls_ on the produced `libasmrun.a` such as `getpid`. I will try to look why.

hannesm:
On my FreeBSd system with 4.13.0 the difference I can observe is that `s.h` (generated by configure, in `ocaml/runtime/caml/s.h`) now `defines HAS_GETCWD 1`, which leads to a different conditionally compiled code in `sys.c`. I consider this behaviour as fine, `caml_sys_getcwd` will still error out. OTOH we may be able to drop the `STUB_ABORT(getcwd)` from nolibc. I can't see any `getuid` refeences.

dinosaure:
Ok I found a final solution, we must keep:
```
        echo '#define HAS_SECURE_GETENV' >> ocaml/runtime/caml/s.h
```

If we delete it, we bring some Unix dependencies. Then, you are right about `GETTIMEOFDAY` and `HAS_TIMES`, the `configure` is able to generate them correctly.

hannesm:
please excuse my lack of Linux testing. I force-pushed to use `-lnolibc -lopenlibm` (instead of `-lopenlibm -lnolibc`) in `Makefile` -- and now this works fine on an Ubuntu 20.04 (focal).

dinosaure:
The `dns-resolver` can compiles :tada:, this PR is the definitive fix for #101 and we should fix #92 with that. Waiting the CI and ready to merge/cut a release.

hannesm:
Sorry, I force-pushed again to use `sed` instead of a million of patches. This makes it more portable (though we've to be careful about new OCaml releases and may at some point need to act differently for different OCaml versions).

dinosaure:
As we said in #101, #104 solves the initial issue on OpenBSD 7. This PR is not needed anymore so. Thanks!

PR #104, opened 2021-12-30
Cross compiler take3

hannesm:
this is #92 rebased on master, and the review I did earlier today is addressed.

I also looked how to integrate #93, but I think this needs some nixos expertise (since now the Makefile rule `ocaml/Makefile` not only unpacks the ocaml-src, but also adapts various files).


hannesm:
The GitHub actions run is at https://github.com/hannesm/ocaml-freestanding/runs/4670324927 (all passed), manually tested that compilation on OpenBSD 7.0 succeeds as well :)

dinosaure:
If someone (@sternenseemann?) can confirm that this PR works on NixOS, we will ready to cut a release 👍 .

configure.sh
```diff
@@ -1,60 +1,110 @@
...
+            usage
+            ;;
+    esac
+
+    shift
+done
+
+[ -z "${CONFIG_TARGET}" ] && die "The --target option needs to be specified."
+
 ocamlfind query ocaml-src >/dev/null || exit 1
```
dinosaure, line 66: May be we should include this patch, from @sternenseemann:
> The avantage of having it only in the Makefile potentially could be that make doesn't try to build it if it is already present.

Even if this is not really needed for the NixOS support.

```diff
- ocaml query ocaml-src >/dev/null || exit 1
+ [ -e "$(dirname "$0")/ocaml" ] || ocamlfind query ocaml-src >/dev/null || exit 1
```

sternenseemann: Is definitely needed for us, since the `ocaml-src` package is not a thing in nixpkgs and it doesn't really have an advantage introducing it.

sternenseemann:
> If someone (@sternenseemann?) can confirm that this PR works on NixOS, we will ready to cut a release +1 .

I've started working on it, but I need to solve the changed solo5 integration first, so it might take a bit of time.

configure.sh
dinosaure: Added via 0bf78a7, thanks!

hannesm:
as mentioned, I doubt that the change in `configure.sh` is sufficient: how does nixos retrieve and put OCaml's source into the `ocaml` subdirectory? What about the changes (`sed -i -e`) that the Makefile rule `ocaml/Makefile` does to the `ocaml/Makefile`? Should these changes instead be part of the `ocaml/Makefile.config` rule so that on nixos, where the `ocaml` subdirectory appears magically, the OCaml configure script is modified to suit ocaml-freestanding needs?

IMHO the ocaml-src check in configure.sh should be removed entirely -- we support installation via opam, and ocaml-src is a dependency, so it will always be present.

I did this in 011cca6 (pushed here), please let me know what you think.

hannesm:
The error is now:
```
Run MODE=virtio opam exec -- dune build --root test
  MODE=virtio opam exec -- dune build --root test
  shell: /usr/bin/bash -e {0}
  env:
    OPAMCLI: 2.0
    OPAMCOLOR: always
    OPAMERRLOGLEN: 0
    OPAMJOBS: 2
    OPAMPRECISETRACKING: 1
    OPAMSOLVERTIMEOUT: 1000
    OPAMVERBOSE: false
    OPAMYES: 1
Entering directory '/home/runner/work/ocaml-freestanding/ocaml-freestanding/test'
Error: Program ocamlfind not found in the tree or in PATH
 (context: freestanding)
Error: Process completed with exit code 1.
```

But I do not understand where `ocamlfind` is executed -- and which package should depend on it? dune? or really ocaml-freestanding?

hannesm:
`ocamlfind` is executed by dune (in `dune_rules/context.ml`).

dinosaure:
> ocamlfind is executed by dune (in dune_rules/context.ml).

It's needed by the `Makefile` too to download OCaml sources.

dinosaure:
From our perspective, this PR seems good for MirageOS 4.0 - the CI about `mirage-skeleton` is green and @hannesm completely deleted the `pkg-config` usage. `ocamlfind` remains (due to `dune`) but that's fine for us. The NixOS support is not done yet but we want to prioritize the MirageOS 4.0 release. Thanks for all of you about review/participation/etc. With Solo5.0.7.0, it's a big milestone.

PR #109, opened 2022-01-23
Add support for OCaml 4.14

kit-ty-kate:
Change made in https://github.com/ocaml/ocaml/commit/c71428e7a3c69f3478c92b8c92918e51aafe311d


dinosaure:
Thanks, it seems good for me :+1:. I don't think that we need a double-check for this PR.

Issue #110, opened 2022-02-28
Floating point issue


palainp:
I'm not really sure if this issue is related to ocaml-freestanding or solo5 or xen or qubes, sorry for the disturbance if I misplaced the issue here.
Actually, I have #MF (floating point exception) error from solo5 with this unikernel (compiled with mirage 4.0~beta3):
```
$ cat unikernel.ml 
open Lwt.Infix

module Hello (Time : Mirage_time.S) = struct
  let start _time =
    Logs.info (fun f -> f "total runtime 1 %.6f s" 0.5);
    Logs.info (fun f -> f "total runtime 2 %.6f s" 1.0);
    Logs.info (fun f -> f "total runtime 3 %.6f s" 0.75);
    Logs.info (fun f -> f "total runtime 4 %.6e s" 0.5);
    Logs.info (fun f -> f "total runtime 5 %.6e s" 0.25);
    Logs.info (fun f -> f "total runtime 6 %.6e s" 0.875);
    Logs.info (fun f -> f "total runtime 7 %.6e s" 0.6);
    Logs.info (fun f -> f "total runtime 8 %.6e s" 0.4);
    Time.sleep_ns (Duration.of_sec 1) >>= fun () ->
    Lwt.return_unit
end
$ cat config.ml 
open Mirage

let main =
  foreign
    "Unikernel.Hello" (time @->job)

let () =
  register "hello" [main $ default_time]
```
It works well with the `x*1/2^y` prints but fails on the first `0.6` print. The faulting address is on a `fldz` instruction ("add 0 to the FPU stack") which is due to something around the compilation of https://github.com/mirage/ocaml-freestanding/blob/7e85ff780deba6f526ff53445b5a0bd4ac6fd86e/nolibc/vfprintf.c#L236
I was not able to recompile ocaml-freestanding with `-msoft-float`.
If needed the kernel can be found there : https://we.tl/t-DEeIN7TW80 (md5sum: 982d75122f9346b0101510c4a5644252)
One strange thing is that it works fine with `Xen 4.8` (Qubes 4.0.4) but not with `Xen 4.14` (Qubes 4.1), I'm not sure if something is broken between these two versions.

palainp:
In addition, compilation and execution with spt work as expected.

dinosaure:
Hope that it will be fixed with #111 but I will let this issue open to let you to recheck with the next release of `ocaml-freestanding` if it's work or not.

palainp:
[...]

PR #111, opened 2022-03-01
vfprintf: change long double to double 

palainp:
I'm not sure this is the best way to fix issue #110 (and as a side effect also fix mirage/qubes-mirage-firewall#134). Basically, it's a trade for the `double long` precision into `double`, which may or may not have an impact when we want to print floating values.

Since the initial problem doesn't seem to be seen outside of Qubes 4.1 (Xen 4.14) and can be caused by something other than that (maybe any change in the defaults between Qubes 4.0.4 and Qubes 4.1 (Xen 4.8 and Xen 4.14 respectively)) and we may have another floating point issue in another piece of code, I'm definitely not sure that lowering the precision here is the way to go  (it's probably better to fix the root cause but I'm now out of ideas for that) but it does have the advantage of allowing unikernel mirage to run with Qubes :)

### DIFF
--- a/nolibc/vfprintf.c
+++ b/nolibc/vfprintf.c
@@ -202,13 +202,13 @@ static char *fmt_u(uintmax_t x, char *s)
 typedef char compiler_defines_long_double_incorrectly[9-(int)sizeof(long double)];
 #endif
 
-static int fmt_fp(FILE *f, long double y, int w, int p, int fl, int t)
+static int fmt_fp(FILE *f, double y, int w, int p, int fl, int t)
 {
-	uint32_t big[(LDBL_MANT_DIG+28)/29 + 1          // mantissa expansion
-		+ (LDBL_MAX_EXP+LDBL_MANT_DIG+28+8)/9]; // exponent expansion
+	uint32_t big[(DBL_MANT_DIG+28)/29 + 1          // mantissa expansion
+		+ (DBL_MAX_EXP+DBL_MANT_DIG+28+8)/9]; // exponent expansion
 	uint32_t *a, *d, *r, *z;
 	int e2=0, e, i, j, l;
-	char buf[9+LDBL_MANT_DIG/4], *s;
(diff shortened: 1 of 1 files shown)


palainp:
Thinking back about the precision drop, and according to https://ocaml.org/manual/values.html, the floating point values are 64bits long (sign bit+53 mantissa bits+10 exponent bits) which will fits into `double` and doesn't seem to require `long double` (the original openlibm may have targets that will take the 16B for `long double` but I think that it's fine here).

Also when I try to force a long double precision print the compiler complains about the print format:
```
File "unikernel.ml", line 10, characters 42-45:
10 |         Logs.info (fun f -> f "hello %Lf" 0.3);
                                               ^^^
Error: This expression has type float but an expression was expected of type
         int
```

hannesm:
> However, with a deep look, I can not find a call of vfprintf in the caml runtime.

In ocaml-freestanding, `vfprintf` is the actual implementation and all derivates (`fprintf`, `printf`, `vsnprintf`) are calling into `vfprintf`.

The OCaml runtime is slightly complex in respect to printf, but on the OCaml side:
- CamlInternalFormat defines a recursive function `make_printf` <- this is called by `Printf.printf`
- This calls out to `make_float_padding_precision`, which calls `convert_float`, which calls `format_float` which is `caml_format_float`
- That is a C implementation in `runtime/floats.c`, which calls `caml_alloc_sprintf` (in `runtime/str.c`), which calls out to `vsnprintf`

I think this PR is nice to merge, though I'm still unsure what happened between Xen 4.8 and Xen 4.14 and whether we'll run in other issues (e.g. with openlibm and trigonometry functions) that are fixed by float/double. I also don't have much knowledge about floating point numbers and precision.

TL;DR: Happy to have this merged since it fixes a concrete issue on Xen.

dinosaure:
Ok, I think I understand more the execution path with what @hannesm said. In that case, I think that it's good to merge too :+1:. Thanks for your work.

PR #113, opened 2022-03-14
Add conf-which as a dependency of ocaml-freestanding

dinosaure:


dinosaure:
That really weird, it seems that in `ubuntu`, the `command -v` does not returns anything where `which` returns the expected result, see this line: https://github.com/mirage/ocaml-freestanding/runs/5552503982?check_suite_focus=true#step:5:5097

dinosaure:
Due to the weird behavior of `command -v`, the best is to add a new build dependency `conf-which`.

PR #114, opened 2022-03-25
Rename freestanding to solo5

dinosaure:
I replaced any relevant occurrence of `freestanding` by `solo5`, it should properly work (build and install) and we need to change `mirage` with such change. I would like a double check however, we can easily miss something important.


samoht:
LGTM - should we also rename the name of the repository? And update the URLs in the opam files?

dinosaure:
>  LGTM - should we also rename the name of the repository? And update the URLs in the opam files?

OPAM files are already updated. I will merge this PR and do the renaming.

dinosaure:
>  LGTM - should we also rename the name of the repository? And update the URLs in the opam files?

OPAM files are already updated. I will merge this PR and do the renaming.

PR #120, opened 2022-04-09
Keep track of allocation and release memory (faster than by calling mallinfo and more accurate than calling footprint)

palainp:
This PR adds a counter to memory usage (simply malloc calls add the size value, and free calls substract the size value). The tricky part is that dmalloc can use `dispose_chunk` to do something similar to free and we have to substract memory used there too.

This doesn't change the API and just adds an exported function `malloc_memory_usage` (I'm not sure that's the best name though).

### DIFF
--- a/nolibc/dlmalloc.i
+++ b/nolibc/dlmalloc.i
@@ -1,4 +1,14 @@
 /*
+  Note: that this is a slightly modified version of dlmalloc to add
+  the internal_memory_usage field to the malloc_state structure. This
+  field is used with the gm variable to keep track of the amount of
+  memory allocated in the heap.
+  The results of this fast estimation are returned by the
+  dlmalloc_memory_usage function exported as malloc_memory_usage.
+  See lines 978, 2619 and in dlmalloc(), dlfree() and dispose_chunk()
+*/
+
+/*
   This is a version (aka dlmalloc) of malloc/free/realloc written by
   Doug Lea and released to the public domain, as explained at
   http://creativecommons.org/publicdomain/zero/1.0/ Send questions,
@@ -827,6 +837,7 @@ extern "C" {
 #define dlmalloc_trim          malloc_trim
 #define dlmalloc_stats         malloc_stats
 #define dlmalloc_usable_size   malloc_usable_size
+#define dlmalloc_memory_usage  malloc_memory_usage
 #define dlmalloc_footprint     malloc_footprint
 #define dlmalloc_max_footprint malloc_max_footprint
 #define dlmalloc_footprint_limit malloc_footprint_limit
@@ -963,6 +974,17 @@ DLMALLOC_EXPORT void* dlvalloc(size_t);
   M_MMAP_THRESHOLD     -3      256*1024   any   (or 0 if no MMAP support)
 */
 DLMALLOC_EXPORT int dlmallopt(int, int);
+
+/*
+  malloc_memory_usage();
+  Returns total number of bytes allocated by malloc, realloc etc. This
+  value is the same as the uordblks field in the mallinfo structure but
+  the value is precomputed value that is updated on each call to malloc,
+  free, etc.
+  Unlike malloc_footprint it does not counts the bytes obtained from the
+  system but not currently allocated.
+*/
+DLMALLOC_EXPORT size_t dlmalloc_memory_usage(void);
 
 /*
   malloc_footprint();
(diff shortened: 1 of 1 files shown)


hannesm:
Thanks for this PR. I was wondering whether the subtraction could be done at a single place, `dispose_chunk`, since it looks like with this PR it happens just before every call site thereof?

nolibc/dlmalloc.i
```diff
@@ -5121,8 +5136,12 @@ static size_t internal_bulk_free(mstate m, void* array[], size_t nelem) {
             set_inuse(m, p, newsize);
             *b = chunk2mem(p);
           }
-          else
+          else {
```
hannesm, line 5139: in this function, `internal_bulk_free`, I don't quite understand why in the above consequence ("try to merge with next chunk") that `internal_memory_usage` is not updated?

palainp: The `set_inuse(m, p, newsize);` marks the bloc `p` to be sized as `newsize`(=current size+next chunk size) and `*b = chunk2mem(p);` will prepare the next loop iteration (`b` is a pointer to the next cell in the array to be freed). We will substract the chunk size only when the next chunk cannot be merged to the current one.

Anyway, by moving the `internal_memory_usage` change into `dispose_chunk` this part of the patch will no longer appear :)

palainp:
> Thanks for this PR. I was wondering whether the subtraction could be done at a single place, `dispose_chunk`, since it looks like with this PR it happens just before every call site thereof?

You're totally right about that, this comes from my tests to find out why the value of `internal_memory_usage` can goes down the real usage and unfortunately I changed two things at once:
* removing static and adding volatile qualifier for `internal_memory_usage`
* the extraction of `internal_memory_usage -= ...` from `dispose_chunk`
But it appears that I now have the substraction before every `dispose_chunk` calls. I'll test out with a single place modification at the begining of `dispose_chunk`.

palainp:
After testing (intensive network usage + 32M unikernel which is barely enough to try to have the worst possible situation) I'm not able to reproduce any difference between the value returned by `mallinfo` and the `internal_memory_usage` value. It's better to declare it as static and with code factorization it is easier to read.

palainp:
I think I need some advice here, looking back to the patch, with regard to the following paragraph 
https://github.com/mirage/ocaml-solo5/blob/80ff24ef8f678e2b2f7c6c8dc45e59caf739a0cb/nolibc/dlmalloc.i#L2990
it may be better to move the `internal_memory_usage` inside the global state:
https://github.com/mirage/ocaml-solo5/blob/80ff24ef8f678e2b2f7c6c8dc45e59caf739a0cb/nolibc/dlmalloc.i#L2587

palainp:
I hope this notice comment will be clear and explanatory enough.

As a note, could it be possible to co-author @winux138 for [e9c5343](https://github.com/mirage/ocaml-solo5/pull/120/commits/e9c53438f7a26a46a4d8354f4a9b3cf42f59b2e1) and [80ff24e](https://github.com/mirage/ocaml-solo5/pull/120/commits/80ff24ef8f678e2b2f7c6c8dc45e59caf739a0cb) who worked on that too ? I'm not sure I have done that correctly.

hannesm:
This looks fine to me. I also looked into the difference of assembly output between main branch and these changes, and there is not much difference (some subs etc., but not any more function calls etc.).

> As a note, could it be possible to co-author @winux138 for [e9c5343](https://github.com/mirage/ocaml-solo5/pull/120/commits/e9c53438f7a26a46a4d8354f4a9b3cf42f59b2e1) and [80ff24e](https://github.com/mirage/ocaml-solo5/pull/120/commits/80ff24ef8f678e2b2f7c6c8dc45e59caf739a0cb) who worked on that too ? I'm not sure I have done that correctly.

You did it at least partially correct, what works is the line in a commit message
`Co-authored-by: NAME <email-address>`

Now, you only put "@winux138" there which relies on GitHub to translate the user ID to a link -- if you put the eMail address there instead, it should work (and be independent of GitHub as a plaform).

Would you also mind to squash all commits into a single one?

nolibc/dlmalloc.i
```diff
@@ -964,6 +975,13 @@ DLMALLOC_EXPORT void* dlvalloc(size_t);
 */
 DLMALLOC_EXPORT int dlmallopt(int, int);
 
+/*
+  var init for malloc_memory_usage
+  We need to declare it as extern and reserve memory for that in the kernel
+  as for sbrk()
```
hannesm, line 981: Could you describe here (a) what does it return (presumably "Returns the number of bytes obtained from the system.") and (b) what's the difference to `mallinfo` and `malloc_footprint`? I'd guess "Unlike mallinfo, this functions returns a precomputed value that is updated on each call to malloc and free. Unlike malloc_footprint ???".

PR #124, opened 2023-01-11
Support of OCaml 5.0 (cleaned version)

dinosaure:
This is the cleaned version of #122. From what I saw, I have some questions:
- It seems that we ~export~ _stub_ `usleep` but we don't implement it. We should probably implement it when Solo5 has a function to sleep (see `solo5_yield`)
- Some functions are just **exported** (like `qsort`) but no implementation are needed - at least, at this level of our usual compilation path. I wonder what happens to these functions. If they are exported, is it because they are required? By whom? For what?

Finally, this PR only offers something that compiles. The project has not yet tested with a unikernel. It should be reconfirmed that qubes-firewall works and if we can get
- dns-resolver (small unikernel)
- unipi (HTTP & Git)
- opam-mirror (block device)

This would be very good to confirm that this PR can be merged and that the MirageOS world can move to OCaml 5.

### DIFF
--- a/nolibc/assert.c
+++ b/nolibc/assert.c
@@ -1,16 +1,10 @@
 #include <stdlib.h>
-#include <string.h>
-#include <unistd.h>
+#include <stdio.h>
 
 /*
  * These functions deliberately do not call printf() or malloc() in order to
  * abort as quickly as possible without triggering further errors.
  */
-
-static void puts(const char *s)
-{
-    (void)write(2, s, strlen(s));
-}
 
 void _assert_fail(const char *file, const char *line, const char *e)
 {
--- a/nolibc/include/signal.h
+++ b/nolibc/include/signal.h
@@ -4,12 +4,19 @@
 typedef int jmp_buf;
 int setjmp(jmp_buf);
 void (*signal(int sig, void (*func)(int)))(int);
-#define SIG_DFL NULL
-#define SIG_IGN NULL
-#define SIG_ERR NULL
+#define SIG_DFL 0
+#define SIG_IGN 0
+#define SIG_ERR 0
+#define SIG_BLOCK 0
+#define SIG_SETMASK 0
 /*
  * The following definitions are not required by the OCaml runtime, but are
  * needed to build the freestanding version of GMP used by Mirage.
+ * For OCaml 5.0.0, it's not totally true. SIG_{BLOCK,SETMASK,IGN,DFL) are
+ * needed by the OCaml runtime.
+ *
+ * NOTE: Solo5 does not implement signals, but we should not trigger
+ * a situation where these values are really used.
  */
 #define SIGFPE 1
 int raise(int);
--- a/nolibc/include/signal.h
+++ b/nolibc/include/signal.h
@@ -21,4 +21,7 @@ void (*signal(int sig, void (*func)(int)))(int);
 #define SIGFPE 1
 int raise(int);
 
+typedef int sigset_t;
+int sigfillset(sigset_t *);
+
 #endif
--- a/nolibc/include/stdatomic.h
+++ b/nolibc/include/stdatomic.h
@@ -1,0 +1,43 @@
+#ifndef _STDATOMIC_H
+#define _STDATOMIC_H
+
+#define atomic_load_explicit(x, mode) ( *x )
+#define atomic_load(x) ( *x )
+
+extern int memory_order_release;
+extern int memory_order_acquire;
+extern int memory_order_relaxed;
+extern int memory_order_seq_cst;
+
+#define atomic_fetch_add(X, Y) ({ __auto_type tmp = *X; *X = tmp + Y; tmp; })
+#define atomic_fetch_add_explicit(X, Y, MOD) atomic_fetch_add(X, Y)
+
+#define atomic_thread_fence(MO) do {} while (0)
+
+typedef unsigned long long atomic_uint_fast64_t;
+
+#define atomic_compare_exchange_strong(OBJ, EXPECTED, DESIRED) \
+  ({ int ret = 0; \
+     if (*OBJ == *EXPECTED) { \
+       *OBJ = DESIRED; \
+       ret = 1; \
+     } \
+     ret; \
+  })
+
+#define atomic_exchange(OBJ, DESIRED) \
+  ({ __auto_type tmp = *OBJ; \
+     *OBJ = DESIRED; \
+     tmp; \
+  })
+
+#define atomic_store(OBJ, DESIRED) do { *OBJ = DESIRED; } while(0)
+#define atomic_store_explicit(OBJ, DESIRED, ORDER) atomic_store(OBJ, DESIRED)
+
+#define atomic_fetch_or(OBJ, ARG) \
+  ({ __auto_type tmp = *OBJ; \
+     *OBJ = *OBJ | ARG; \
+     tmp; \
+  })
+
+#endif
--- a/nolibc/include/stdatomic.h
+++ b/nolibc/include/stdatomic.h
@@ -1,6 +1,11 @@
 #ifndef _STDATOMIC_H
 #define _STDATOMIC_H
 
+// Compatibility with non-clang compilers
+#ifndef __has_builtin
+#  define __has_builtin(x) 0
+#endif
+
 #define atomic_load_explicit(x, mode) ( *x )
 #define atomic_load(x) ( *x )
 
@@ -9,13 +14,23 @@ extern int memory_order_acquire;
 extern int memory_order_relaxed;
 extern int memory_order_seq_cst;
 
+#if __has_builtin(__c11_atomic_fetch_add)
+#define atomic_fetch_add(X, Y) __c11_atomic_fetch_add(X, Y, __ATOMIC_SEQ_CST)
+#else
 #define atomic_fetch_add(X, Y) ({ __auto_type tmp = *X; *X = tmp + Y; tmp; })
+#endif
+
 #define atomic_fetch_add_explicit(X, Y, MOD) atomic_fetch_add(X, Y)
 
 #define atomic_thread_fence(MO) do {} while (0)
 
 typedef unsigned long long atomic_uint_fast64_t;
 
+#if __has_builtin(__c11_atomic_compare_exchange_strong)
+#define atomic_compare_exchange_strong(OBJ, EXPECTED, DESIRED) \
+	__c11_atomic_compare_exchange_strong(OBJ, EXPECTED, DESIRED, \
+			__ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)
+#else
 #define atomic_compare_exchange_strong(OBJ, EXPECTED, DESIRED) \
   ({ int ret = 0; \
      if (*OBJ == *EXPECTED) { \
@@ -24,20 +39,31 @@ typedef unsigned long long atomic_uint_fast64_t;
      } \
      ret; \
   })
+#endif
 
+#if __has_builtin(__c11_atomic_exchange)
+#define atomic_exchange(OBJ, DESIRED) \
+	__c11_atomic_exchange(OBJ, DESIRED, __ATOMIC_SEQ_CST)
+#else
 #define atomic_exchange(OBJ, DESIRED) \
   ({ __auto_type tmp = *OBJ; \
      *OBJ = DESIRED; \
      tmp; \
   })
+#endif
 
 #define atomic_store(OBJ, DESIRED) do { *OBJ = DESIRED; } while(0)
 #define atomic_store_explicit(OBJ, DESIRED, ORDER) atomic_store(OBJ, DESIRED)
 
+#if __has_builtin(__c11_atomic_fetch_or)
+#define atomic_fetch_or(OBJ, ARG) \
+	__c11_atomic_fetch_or(OBJ, ARG, __ATOMIC_SEQ_CST)
+#else
 #define atomic_fetch_or(OBJ, ARG) \
   ({ __auto_type tmp = *OBJ; \
      *OBJ = *OBJ | ARG; \
      tmp; \
   })
+#endif
 
 #endif
--- a/nolibc/include/stdio.h
+++ b/nolibc/include/stdio.h
@@ -33,5 +33,10 @@ size_t fwrite(const void *, size_t, size_t, FILE *);
 int fputc(int, FILE *);
 int putc(int, FILE *);
 int ferror(FILE *);
+int fputs(const char *, FILE *);
+FILE *fopen(const char *, const char *);
+int fclose(FILE *);
+int puts(const char *);
+int putchar(int);
 
 #endif
--- a/nolibc/include/sys/mman.h
+++ b/nolibc/include/sys/mman.h
@@ -1,0 +1,25 @@
+#ifndef _MMAP_H
+#define _MMAP_H
+
+#include <stddef.h>
+#include <sys/types.h>
+
+void *mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off);
+
+#define PROT_NONE 0
+#define PROT_READ 1
+#define PROT_WRITE 2
+
+#define MAP_SHARED 0x01
+#define MAP_PRIVATE 0x02
+#define MAP_FIXED 0x10
+#define MAP_ANONYMOUS 0x20
+#define MAP_ANON MAP_ANONYMOUS
+
+#define MAP_FAILED NULL
+
+#define OCAML_SOLO5_PAGESIZE (1 << 12)
+
+int munmap(void *addr, size_t len);
+
+#endif
--- a/nolibc/include/sys/stat.h
+++ b/nolibc/include/sys/stat.h
@@ -11,6 +11,8 @@ struct stat {
 #define S_IFMT 0
 #define S_IFREG 0
 #define S_ISREG(x) (0)
+#define S_IRUSR 0
+#define S_IWUSR 0
 int stat(const char *, struct stat *);
 int mkdir(const char *, mode_t);
 
--- a/nolibc/include/sys/stat.h
+++ b/nolibc/include/sys/stat.h
@@ -15,5 +15,6 @@ struct stat {
 #define S_IWUSR 0
 int stat(const char *, struct stat *);
 int mkdir(const char *, mode_t);
+int fstat(int, struct stat *);
 
 #endif
--- a/nolibc/mmap.c
+++ b/nolibc/mmap.c
@@ -1,0 +1,35 @@
+#include <stdlib.h>
+#include <stdio.h>
+
+#include <sys/mman.h>
+
+void *mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off) {
+
+  /* man page for mmap says:
+   * If addr is not NULL, then the kernel takes it as a hint about where to place
+   * the mapping; [...] If another apping already exists there, the kernel picks
+   * a new address that may or *may not* depend on the hint.
+   *
+   * XXX(dinosaure): for our purpose (Solo5 & OCaml), OCaml does not require a
+   * specific (aligned) address from [mmap]. We can use [malloc()] instead of.
+   * The OCaml usage of [mmap()] is only to allocate some spaces, only [fildes
+   * == -1] is handled so.
+   */
+  (void)addr;
+  (void)prot;
+  if (fildes != -1) {
+    printf("mmap: file descriptor is unsupported.\n");
+    abort();
+  }
+  if (!(flags & MAP_ANONYMOUS) || off != 0) {
+    printf("mmap: only MAP_ANONYMOUS (and offset is 0) is supported.\n");
+    abort();
+  }
+
+  void *ptr = NULL;
+  posix_memalign(&ptr, OCAML_SOLO5_PAGESIZE, len);
+  /* XXX(palainp): Solo5 returns -1 and set errno on error, and it does not
+   * modify ptr, however both are not standardized. We can return NULL to the
+   * caller on error and do not check the returned value here. */
+  return ptr;
+}
--- a/nolibc/puts.c
+++ b/nolibc/puts.c
@@ -1,0 +1,17 @@
+#include <string.h>
+#include <unistd.h>
+
+extern void solo5_console_write(const char *, size_t);
+
+int puts(const char *s)
+{
+    size_t len = strlen(s);
+    solo5_console_write(s, len);
+    return (len);
+}
+
+int putchar(int chr)
+{
+    solo5_console_write((char *) &chr, 1);
+    return (1);
+}
--- a/nolibc/stubs.c
+++ b/nolibc/stubs.c
@@ -78,6 +78,7 @@ STUB_ABORT(read);
 STUB_IGNORE(int, readlink, -1);
 STUB_ABORT(unlink);
 STUB_ABORT(rmdir);
+STUB_ABORT(ftruncate);
 
 /* dirent.h */
 STUB_WARN_ONCE(int, closedir, -1);
--- a/nolibc/stubs.c
+++ b/nolibc/stubs.c
@@ -147,3 +147,4 @@ STUB_ABORT(pthread_self);
 STUB_ABORT(pthread_detach);
 
 STUB_ABORT(sigfillset);
+STUB_ABORT(strerror_r);
--- a/nolibc/stubs.c
+++ b/nolibc/stubs.c
@@ -100,3 +100,45 @@ STUB_ABORT(strerror);
 /* sys/stat.h */
 STUB_WARN_ONCE(int, stat, -1);
 STUB_ABORT(mkdir);
+
+/* pthread.h */
+STUB_IGNORE(int, pthread_join, 0);
+STUB_IGNORE(int, pthread_create, 0);
+STUB_IGNORE(int, pthread_attr_init, 0);
+STUB_ABORT(pthread_cleanup_push);
+STUB_ABORT(pthread_cleanup_pop);
+
+STUB_IGNORE(int, munmap, 0); /* TODO: Implement munmap correctly and avoid memory leaks */
+
+int memory_order_release;
+int memory_order_acquire;
+int memory_order_relaxed;
+
+STUB_ABORT(atomic_store_explicit);
+STUB_ABORT(atomic_exchange);
+
+/* above that line, for OCaml 5, those are only required (i guess) for the configure step */
+STUB_IGNORE(int, pthread_mutex_lock, 0);
+STUB_IGNORE(int, pthread_mutex_trylock, 0);
+STUB_IGNORE(int, pthread_mutex_unlock, 0);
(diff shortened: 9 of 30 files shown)


haesbaert:
IMHO this is far from ready, we didn't really review much, we just got it to work.

dinosaure:
A precision about words used in my message:
- `stub` is when we actually provide an implementation which raises an error
- `export` is when we just describe the function (no implementation is done here, only `strcpy` is a special case when it is provided by Solo5)
- `define` is about values

nolibc/sysdeps_solo5.c
```diff
@@ -76,6 +76,7 @@ static uintptr_t sbrk_start;
 static uintptr_t sbrk_end;
 static uintptr_t sbrk_cur;
 static uintptr_t sbrk_guard_size;
+static uint8_t   tls[4096];
```
palainp, line 79: The hardcoded 4k size should be set to something sensible and automatically adjusted to what the runtime needs, but I don't know yet how to have the correct value.

nolibc/sysconf.c
```diff
@@ -0,0 +1,10 @@
+#include <unistd.h>
+
+long sysconf(int x) {
+  switch (x) {
+  case _SC_PAGESIZE:
```
palainp, line 5: Also add the other synonym `_SC_PAGE_SIZE`?

nolibc/stubs.c
```diff
@@ -100,3 +104,49 @@ STUB_ABORT(strerror);
...
 STUB_ABORT(mkdir);
+
+/* pthread.h */
+STUB_IGNORE(int, pthread_join, 0);
+STUB_IGNORE(int, pthread_create, 0);
+STUB_IGNORE(int, pthread_attr_init, 0);
+STUB_ABORT(pthread_cleanup_push);
+STUB_ABORT(pthread_cleanup_pop);
+
+STUB_IGNORE(int, munmap, 0); /* TODO: Implement munmap correctly and avoid memory leaks */
```
palainp, line 115: I think this is not trivial or impossible with the current dlmalloc implementation. The best tool we can have is `realloc_in_place` which will probably have troubles to find the correct memory chunk as the address can be something inside an allocated chunk. i.e.
```C
    int pagesize = sysconf(_SC_PAGESIZE);
    char *addr = mmap(NULL, 4 * pagesize, PROT_NONE, MAP_PRIVATE, 0);
    munmap(addr+pagesize, pagesize); /* now we have 4 pages allocated and 1 in the middle should be freed...*/
```

nolibc/include/stdio.h
```diff
@@ -33,5 +33,16 @@ size_t fwrite(const void *, size_t, size_t, FILE *);
...
 int ferror(FILE *);
+int fputs(const char *, FILE *);
+FILE *fopen(const char *, const char *);
+int fclose(FILE *);
+
+/* XXX(dinosaure): currently, these functions are **not**
+ * stubed. They exists into our [nolibc] or Solo5:
+ * - [strcpy] exists into [bindings/lib.c] from Solo5
+ * - [puts]/[putchar] are provided by [puts.c] */
+char *strcpy(char *, const char *);
```
palainp, line 44: `strccpy` shouldn't be in `string.h`? and its implementation can probably call `strncpy`?

nolibc/mmap.c
```diff
@@ -0,0 +1,27 @@
...
+   */
+  (void)addr;
+  if (fildes != -1) {
+    printf("mmap: file descriptor is unsupported.\n");
+    abort();
+  }
+  (void)prot;
+  (void)flags;
+  (void)off;
+  return malloc(len);
```
palainp, line 26: As the address should be page-aligned, it can probably be better to call:
```C
    return posix_memalign(PAGESIZE, len);
```
`PAGESIZE` is the result of `sysconf(_SC_PAGESIZE)` but, regarding the comment in `sysconf.c`, maybe it can be better to `#define PAGESIZE 1<<12` somewhere?

nolibc/sysconf.c
```diff
@@ -0,0 +1,10 @@
...
+    return (1 << 12); /* TODO: How do we do better? */
```
palainp, line 6: related to `mmap` maybe we can define a PAGESIZE macro somewhere?

nolibc/assert.c
```diff
@@ -1,16 +1,15 @@
...
 /*
  * These functions deliberately do not call printf() or malloc() in order to
  * abort as quickly as possible without triggering further errors.
  */
 
-static void puts(const char *s)
-{
-    (void)write(2, s, strlen(s));
-}
+// static void puts(const char *s)
```
palainp, line 9: Maybe we can remove that unused anymore code instead of commenting it out?

nolibc/stubs.c
palainp: There is some references to `munmap` in `ocaml/ocaml/runtime`, one that may occurs during the life of an unikernel can be https://github.com/ocaml/ocaml/blob/cbfa40a606354aba9aa81aefb5938d5c4b3982fe/otherlibs/runtime_events/runtime_events_consumer.c#L290 (not sure what a cursor can be).

dinosaure: It seems related to the runtime event ringbuffer:
https://github.com/ocaml/ocaml/blob/34cf5aafcedc2f7895c7f5f0ac27c7e58e4f4adf/otherlibs/runtime_events/runtime_events.mli#L186-L198

That mostly means that the current way to manipulate the runtime event ringbuffer _via_ OCaml produces a memory leak. However, it seems that this `munmap` is related to a `mmap` on a file-descriptor:
https://github.com/ocaml/ocaml/blob/cbfa40a606354aba9aa81aefb5938d5c4b3982fe/otherlibs/runtime_events/runtime_events_consumer.c#L199-L201

So, if it's the only reference of `munmap`, I would like to say that it's ~ fine ~. In anyway, we are not able to manipulate file-descriptor with Solo5.

nolibc/sysdeps_solo5.c
haesbaert: what happens if it needs more ? is it detected or just overwrites memory ?

nolibc/mmap.c
palainp: related to the discussion about `munmap`, this is a possible implementation that checks for the `MAP_ANONYMOUS` flag (and the PR works with mirage-qubes-firewall):
```C
void *mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off)
{

  /* man page for mmap says:
   * If addr is not NULL, then the kernel takes it as a hint about where to place
   * the mapping; [...] If another apping already exists there, the kernel picks
   * a new address that may or *may not* depend on the hint.
   *
   * The OCaml usage of [mmap()] is only to allocate some spaces, only [fildes
   * == -1] and [MAP_ANONYMOUS with offset == 0] are handled so.
   */
  (void)addr;
  (void)prot;

  if (fildes != -1) {
    printf("mmap: file descriptor is unsupported.\n");
    abort();
  }
  if (!(flags & MAP_ANONYMOUS) || off != 0)
  {
    printf("mmap: only MAP_ANONYMOUS (and offset is 0) is supported.\n");
    abort();
  }

  void *ptr;
  posix_memalign(&ptr, PAGESIZE, len);
  return ptr;
}
```

nolibc/include/sys/mman.h
```diff
@@ -0,0 +1,23 @@
...
+
+typedef int off_t;
+
+void *mmap(void *addr, size_t len, int prot, int flags, int fildes, off_t off);
+
+#define PROT_NONE 0
+#define PROT_READ 1
+#define PROT_WRITE 2
+
+#define MAP_PRIVATE 0
```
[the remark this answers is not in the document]
haesbaert: also off_t should not be int, better check what solo5 is using for similar things, but  should be the size of size_t (probably it's a long)

nolibc/stubs.c
palainp: I agree and the `MAP_ANONYMOUS` seems so far to be respected.

nolibc/sysdeps_solo5.c
palainp: It does not seem to be detected at all (I just tried with 1B for TLS). So far I think it'll overwrites memory and I'll get troubles at some point (unsure what it currently overwrites ;) ).

haesbaert: That's like.. hmm that's a bummer, man.
I have to read on TLS but ideally we would have a guard-page after TLS to make sure it blows up, we have to check what the others do.

dinosaure: I think we need a confirmation of the OCaml core team about that (/cc @Engil).

dinosaure: Yes, but they probably assume a certain size :) again, we must be on the perspective that we tune every C details for the OCaml runtime (as `mmap`).

haesbaert: I doubt they calculated, at any rate we should be able to blowup if it runs over, but I think that is impossible in solo5 because we can't have a guard-page.

nolibc/include/sys/mman.h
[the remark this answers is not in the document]
palainp: It seems to be `uint64_t` in Solo5 (https://github.com/Solo5/solo5/blob/85ce3239ce87a6e49e550a7a0756b072a8c9fa6c/include/solo5.h#L257) so we could probably define it to `size_t` (or `uin64_t` but it has the semantics of a size).

nolibc/include/stdio.h
hannesm: but who's in need of `strcpy`? I thought the OCaml runtime would not use these unsafe C functions anymore...

nolibc/sysdeps_solo5.c
hannesm: we can adapt solo5 to reserve a specific memory area for tls, and have this region being guarded. of course that'd mean to introduce complexity into solo5...

hannesm:
From the cirrus CI, there is one warning and one error:
```
Entering directory '/tmp/cirrus-ci-build/test'
(cd _build/solo5 && /.opam/5.0.0/bin/x86_64-solo5-none-static-cc -I/.opam/5.0.0/solo5-sysroot/include/nolibc/ -include _solo5/overrides.h -O2 -fno-strict-aliasing -fwrapv -pthread -D_FILE_OFFSET_BITS=64 -I/.opam/5.0.0/.opam-switch/build/ocaml-solo5.0.8.1/nolibc/include -include _solo5/overrides.h -I/.opam/5.0.0/.opam-switch/build/ocaml-solo5.0.8.1/openlibm/include -I/.opam/5.0.0/.opam-switch/build/ocaml-solo5.0.8.1/openlibm/src -nostdlib -I/.opam/5.0.0/solo5-sysroot/include/nolibc/ -include _solo5/overrides.h -O2 -fno-strict-aliasing -fwrapv -pthread -g -I /.opam/5.0.0/solo5-sysroot/lib/ocaml -o startup.o -c startup.c)
startup.c:16:1: warning: control reaches end of non-void function [-Wreturn-type]
}
^
1 warning generated.
File "dune", line 2, characters 7-11:
2 |  (name main)
           ^^^^
(cd _build/solo5 && /.opam/5.0.0/solo5-sysroot/bin/ocamlopt -w @1..3@5..28@30..39@43@46..47@49..57@61..62-40 -strict-sequence -strict-formats -short-paths -keep-locs -g -o main.exe manifest.o startup.o .main.eobjs/native/dune__exe__Main.cmx -g -w +A-4-41-42-44 -bin-annot -strict-sequence -principal -safe-string -color always -cclib '-z solo5-abi=hvt')
ld.lld: error: /.opam/5.0.0/solo5-sysroot/lib/ocaml/libasmrun.a(domain.n.o) has an STT_TLS symbol but doesn't have an SHF_TLS section
cc: error: linker command failed with exit code 1 (use -v to see invocation)
File "caml_startup", line 1:
Error: Error during linking (exit code 1)
```

Would be great if someone more into TLS could take a look at the TLS issue (I suspect this is due to usage of clang/llvm both as compiler and as linker).

palainp:
>  `startup.c:16:1: warning: control reaches end of non-void function [-Wreturn-type]`

Indeed the `solo5_app_main` (in `test/startup.c` and also `example/startup.c`) does not return anything (and probably never reach that point), `mirage/mirage-xen` returns 0, and the tests suites in Solo5 return `SOLO5_EXIT_SUCCESS` or `SOLO5_EXIT_FAILURE`. I guess `return 0;` should be fine here.

> ld.lld: error: /.opam/5.0.0/solo5-sysroot/lib/ocaml/libasmrun.a(domain.n.o) has an STT_TLS symbol but doesn't have an SHF_TLS section

It's interesting that the error was already spotted in Solo5 in https://github.com/Solo5/solo5/blob/7ba74eb9f645e07570cd35a833eddec4918eaf64/tests/test_tls/test_tls.c. And the test is deactivated when compiled with lld. The check in lld seems to be https://github.com/llvm-mirror/lld/blob/64b024a57c56c3528d6be3d14be5e3da42614a6f/ELF/Symbols.cpp#L112.

dinosaure:
Thanks for your feedback all, I integrated most of what you said, only the TLS problem remains now.

nolibc/puts.c
```diff
@@ -0,0 +1,17 @@
+#include <string.h>
+#include <unistd.h>
+
+extern void solo5_console_write(const char *, size_t);
+
+int puts(const char *s)
+{
+    int len = strlen(s);
```
palainp, line 8: `strlen` returns a `size_t` so `len` should.

hannesm:
Thanks for all the work @palainp @kit-ty-kate @dinosaure @haesbaert

About the thread-local storage, I see some issues:
- as @haesbaert said the prefered failure behaviour would be to fail loudly or work (i.e. no silent failure / overwrite of memory)
- to get to that point, I suspect we'll need to understand how the OCaml allocates stuff in the thread local storage area (but additionally, arbitrary C bindings may use `__thread` variable as well)
- the linker error with llvm is something we should address (as pointed out by @palainp this is likely work needed in solo5 (adjusting linker scripts - see https://github.com/Solo5/solo5/pull/436 and https://github.com/Solo5/solo5/pull/420)

I started to read https://maskray.me/blog/2021-02-14-all-about-thread-local-storage to get a better understanding of thread local storage, but won't have much time this month to work on a more thorough review or patches.

nolibc/include/stdatomic.h
```diff
@@ -9,13 +14,22 @@ extern int memory_order_acquire;
...
+
 #define atomic_fetch_add_explicit(X, Y, MOD) atomic_fetch_add(X, Y)
 
 #define atomic_thread_fence(MO) do {} while (0)
 
 typedef unsigned long long atomic_uint_fast64_t;
 
+#if __has_builtin(__c11_atomic_compare_exchange_strong)
+#define atomic_compare_exchange_strong(OBJ, EXPECTED, DESIRED) \
+	__c11_atomic_compare_exchange_strong(OBJ, EXPECTED, DESIRED)
```
palainp, line 31: It should be `___c11_atomic_compare_exchange_strong(OBJ, EXPECTED, DESIRED, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST)`

nolibc/include/stdatomic.h
```diff
@@ -24,20 +38,31 @@ typedef unsigned long long atomic_uint_fast64_t;
      } \
      ret; \
   })
+#endif
 
+#if __has_builtin(__c11_atomic_exchange)
+#define atomic_exchange(OBJ, DESIRED) \
+	__c11_atomic_exchange(OBJ, DESIRED)
```
palainp, line 45: Same here, it should be `__c11_atomic_exchange(OBJ, DESIRED, __ATOMIC_SEQ_CST)`

nolibc/include/stdatomic.h
```diff
@@ -24,20 +38,31 @@ typedef unsigned long long atomic_uint_fast64_t;
...
      tmp; \
   })
+#endif
 
 #define atomic_store(OBJ, DESIRED) do { *OBJ = DESIRED; } while(0)
 #define atomic_store_explicit(OBJ, DESIRED, ORDER) atomic_store(OBJ, DESIRED)
 
+#if __has_builtin(__c11_atomic_fetch_or)
+#define atomic_fetch_or(OBJ, ARG) \
+	__c11_atomic_fetch_or(OBJ, ARG)
```
palainp, line 59: And also here `__c11_atomic_fetch_or(OBJ, ARG, __ATOMIC_SEQ_CST)`

palainp:
Along with the solo5 PR (https://github.com/Solo5/solo5/pull/542), I tried to correctly use the TLS variant II for my processor (https://github.com/Solo5/solo5/blob/8c3a744f998b9977cfc6cd29e0cc40ae2efba167/tests/test_tls/test_tls.c#L52 TLS point to an address after the area) on OpenBSD. And it works (at least a few steps further)!

However I still have a page fault error:
```            |      ___|                
  __|  _ \  |  _ \ __ \
\__ \ (   | | (   |  ) |
____/\___/ _|\___/____/
Solo5: Bindings version v0.7.1-66-gf1daea2
Solo5: Memory map: 512 MB addressable:
Solo5:   reserved @ (0x0 - 0xfffff)
Solo5:       text @ (0x100000 - 0x21bfff)
Solo5:     rodata @ (0x21c000 - 0x22cfff)
Solo5:       data @ (0x22d000 - 0x2e3fff)
Solo5:       heap >= 0x2e4000 < stack < 0x20000000
Solo5: trap: type=#PF ec=0x0 rip=0x1f8f70 rsp=0x1ffffe98 rflags=0x10002
Solo5: trap: cr2=0x0
Solo5: ABORT: cpu_x86_64.c:181: Fatal trap
```
The `0x1f8f70` address is something really strange here:
```
00000000001f8f70 <caml_alloc_final_info>:
  1f8f70:       4c 8b 1d 89 70 e0 ff    mov    -2068343(%rip),%r11        # 0 <domain_field_caml_young_limit>
  1f8f77:       4c 33 1c 24             xor    (%rsp),%r11
  1f8f7b:       55                      push   %rbp
  1f8f7c:       48 89 e5                mov    %rsp,%rbp
  1f8f7f:       41 53                   push   %r11
  1f8f81:       41 56                   push   %r14
  1f8f83:       bf 70 00 00 00          mov    $0x70,%edi
  1f8f88:       e8 b3 22 01 00          callq  20b240 <caml_stat_alloc_noexc>
  1f8f8d:       49 89 c6                mov    %rax,%r14
  1f8f90:       48 85 c0                test   %rax,%rax
  1f8f93:       74 0f                   je     1f8fa4 <caml_alloc_final_info+0x34>
  1f8f95:       ba 70 00 00 00          mov    $0x70,%edx
  1f8f9a:       4c 89 f7                mov    %r14,%rdi
  1f8f9d:       31 f6                   xor    %esi,%esi
  1f8f9f:       e8 7c 3b f8 ff          callq  17cb20 <memset>
...
```
The top instruction is related to stack protection and for all other function I inspected it uses a `retguard` symbol (as example the previous function in the same C file):
```
00000000001f8f20 <caml_final_release>:
  1f8f20:       4c 8b 1d c1 2f 0e 00    mov    929729(%rip),%r11        # 2dbee8 <__retguard_3085>
  1f8f27:       4c 33 1c 24             xor    (%rsp),%r11
  1f8f2b:       55                      push   %rbp
  1f8f2c:       48 89 e5                mov    %rsp,%rbp
...
```
I don't understand why for that specific function the behavior is changed :(

nolibc/sysdeps_solo5.c
palainp: Resolved in the latest Solo5 release

hannesm:
I'm curious, now that solo5 0.8 hit the release road, what is the status of this PR? Does it work on some (all?) platforms? Should we proceed (with review, more testing) in order to get MirageOS working with OCaml 5?

A brief message would be welcome, maybe @palainp knows the details (since he was involved in the most recent commits).

Thanks a lot.

TheLortex:
Hi, thank you again for this work ! 
I managed to build `eio-solo5` on top of that and assemble the network stack up to dream.(https://github.com/TheLortex/eio-solo5) 

Using this proof of concept I've been able to test solo5-hvt with ocaml 5 / eio and didn't encounter any issue.

hannesm:
> Hi, thank you again for this work ! I managed to build `eio-solo5` on top of that and assemble the network stack up to dream.(https://github.com/TheLortex/eio-solo5)
> 
> Using this proof of concept I've been able to test solo5-hvt with ocaml 5 / eio and didn't encounter any issue.

Great to hear. From my point of view, we should first review this PR, test it thoroughly and release it (without switching the effective layer in MirageOS), also evaluate performance etc.

I understand that developing the cool new stuff is exciting, but if nobody cares about upstreaming and pushing the PR to something that we're happy to maintain, there'll be all these vendoring / pin to random branches (that are a nightmare to update).

So, TL;DR: my approach is still "upstream first". Since I'm a bit short on time (and my OCaml 5 knowledge is barely existant), I asked whether there are others (including you :) that would go through the code and thoroughly review it. :)

ocaml-solo5.opam
```diff
@@ -22,7 +22,7 @@ depends: [
   "conf-which" {build}
   "ocamlfind" {build} # needed by dune context (for tests)
   "ocaml-src" {build}
-  "ocaml" {>= "4.12.1" & < "4.15.0"}
+  "ocaml" {>= "5.0" & < "5.1"}
   "solo5" {>= "0.7.0"}
```
palainp, line 26: Should be `  "solo5" {>= "0.8.0"}`.

ocaml-solo5-cross-aarch64.opam
```diff
@@ -21,7 +21,7 @@ depends: [
   "conf-which" {build}
   "ocamlfind" {build} # needed by dune context (for tests)
   "ocaml-src" {build}
-  "ocaml" {>= "4.12.1" & < "4.15.0"}
+  "ocaml" {>= "5.0" & < "5.1"}
   "solo5" {>= "0.7.0"}
```
palainp, line 25: Should be `  "solo5" {>= "0.8.0"}`.

nolibc/include/signal.h
```diff
@@ -4,14 +4,24 @@
...
+#define SIG_DFL 0
+#define SIG_IGN 0
+#define SIG_ERR 0
+#define SIG_BLOCK 0
+#define SIG_SETMASK 0
 /*
  * The following definitions are not required by the OCaml runtime, but are
  * needed to build the freestanding version of GMP used by Mirage.
+ * For OCaml 5.0.0, it's not totally true. SIG_{BLOCK,SETMASK,IGN,DFL) are
+ * needed by the OCaml runtime.
```
fabbing, line 16:
```suggestion
 * The following definitions are required to build the freestanding version of GMP used by Mirage.
OCaml 5.0.0 also required the SIG_{BLOCK,SETMASK,IGN,DFL) definition.
```
We rephrased part of the comment that was outdated by OCaml 5.0.0 support.

nolibc/include/sys/stat.h
```diff
@@ -11,7 +11,10 @@ struct stat {
 #define S_IFMT 0
 #define S_IFREG 0
 #define S_ISREG(x) (0)
```
fabbing, line 13:
```suggestion
#define S_ISREG(x) 0
```
Nitpick, the parentheses were not useful here.

fabbing (commented):
@shym and I have reviewed this PR and it looks good to us, with a few minor suggestions. Although we only tested it with toy examples.

We also have a couple of questions:
- Why not use only one `sed` command per file in the `Makefile` to improve readability and to avoid multiple command executions and temporary files?
- Shouldn't a comment be added to `nolibc/stubs.c` to explain the choice between preferring one of `STUB_ABORT`, `STUB_IGNORE`, `STUB_WARN_ONCE`?

We think it would be interesting to target OCaml 5.1.1 directly, as it brings bug and performance fixes. The build system has also changed a lot since then, which would make it difficult to support both 5.0.0 and 5.1.1.
We are considering working on this, but would be eager to hear your opinion before starting.

palainp:
@fabbing @shym, thank you for the review!

I'm only speaking for myself, but I totally approve if the building could be reworked (even if it doesn't work with 5.0.0, as it doesn't seem possible to deal with both 4.14.1 and 5+, it doesn't matter if 5.0 is skipped in the process).

One thing that still needs work is the current absence of `munmap`  (currently `STUB_IGNORE`) which leads to memory leaks (probably only visible with bigger tests than the toy examples). So far, I've written code for that which needs to be reviewed and tested in the pending #129 PR to this PR :sweat_smile: or in #131 for the 4.14 backport.

dinosaure:
> Why not use only one sed command per file in the Makefile to improve readability and to avoid multiple command executions and temporary files?

The advantage of having several seds is that you can contextualise them in relation to their commits and the PRs that have integrated them. These modifications can be difficult to understand in their current form, and historical information helps us to better understand the ins and outs of these arbitrary choices.

> Shouldn't a comment be added to nolibc/stubs‧c to explain the choice between preferring one of STUB_ABORT, STUB_IGNORE, STUB_WARN_ONCE?

More comments is better :+1:.

> We think it would be interesting to target OCaml 5.1.1 directly, as it brings bug and performance fixes. The build system has also changed a lot since then, which would make it difficult to support both 5.0.0 and 5.1.1.

This PR is always set aside because these versions (5.0 and 5.1) lack compaction. The unikernels we are developing are mainly services. We therefore need a runtime capable of using memory that can be defragmented (and that the GC compacts) in order to provide these services for as long as possible.

As @palainp pointed out, we may need to test allocation logic other than `dlmalloc`. Here again, we need to do some testing (in particular with unikernels such as `dns-resolver` or `tlstunnel`) with such an allocator to see whether it corresponds to our use or not.

We will be delighted to have an extension of this PR with OCaml 5.2. And I'd be delighted to see more details of a possibility to experiment with OCaml 5 while keeping OCaml 4.14 support (I'll be more available for February).

hannesm:
Dear @fabbing @shym,

your review is highly welcome. I do agree with @dinosaure that likely we should target OCaml 5.2 (which allows compaction), and skip 5.0, 5.1 series.

What I'm curious about is that we have some review comments in https://github.com/mirage/ocaml-solo5/pull/122 (or are they all addressed?) -- would be great if you could take a look at them (and comment those which are resolved, maybe re-raise them here if they are not yet resolved).

And then there is https://github.com/mirage/ocaml-solo5/pull/129 -- If I understand correctly, that's mostly additional mmap stuff. In the end, it would be nice to agree on a single PR (this one?), and put the mmap changes here as well.

In terms of ocaml-solo5 and OCaml 4 / OCaml 5 support: I don't think it is worth to have ocaml-solo5 that supports both versions of OCaml at the same time, so we should move the OCaml 4 support to a dedicated branch (in order to keep it up to date when OCaml upstream changes), and support only OCaml 5 in the main branch.

hannesm:
This has been superseeded by #134, closing. Please let me know if this is not the case. The review comments have been addressed in #134 as far as I know. But obviously, another round of reviews or triaging the comments here and in #122 whether they're still applicable to #134 won't hurt.

PR #125, opened 2023-01-19
500 cleaned patch

palainp:
This uses the builtin clang macro for some atomic related function (I got some troubles compiling with clang without that).

It also fix a warning `control reaches end of non-void function [-Wreturn-type]` in the test file.


dinosaure:
Inserted into the initial PR, thanks :+1:.

PR #128, opened 2023-12-08
Fix the compilation of the OCaml toolchain and specify the right target value according to how we want to compile with solo5

dinosaure:
Spotted by `bob` and `esperanto`, we currently don't really specify the right target (at least, the architecture of it) which is needed by some packages like `ocaml-gmp` to be able to cross-compile artifacts to the right assembler. This patch want to clarify that and allows us to really cross-compile a project to `aarch64`. It was not spotted because we never tried to use `ocaml-solo5` as a real cross-compiler for `aarch64` for larger projects.


hannesm:
Looks fine to me. Should we eventually disable the "ocaml-ci" in this repository, since it doesn't work anyways?

PR #134, opened 2024-04-18
Support OCaml 5.2

shym:
This PR brings support for OCaml 5.2 (5.2.0~beta2 at the moment).
It proposes to do so by revising substantially how the cross compiler is built and so, unfortunately, drops support for OCaml 4.x compilers.

This builds upon the great work that has been done in the following previous PRs:

- #122
- #124
- #129

and so I chose to branch it from #129’s branch, before the changes to the memory allocator (which are not necessary for OCaml 5.2), and rebased it all on `main`. All the comments in those PRs have been taken into account.

This PR is marked as draft as it should be merged only when OCaml 5.2 is released but it can be tested in a 5.2.0~beta2 switch.

## Main changes

To support OCaml 5.2, this PR extends `nolibc` as necessary to build the runtime. Most of those extensions are inherited from those previous PRs.
Apart from that, the main changes of this PR are as follows.

### Patches to the compiler sources

This PR proposes to change from the scheme of modifying the OCaml build system by `sed`s and `echo`s and instead brings a series of (version-specific) patches in `patches/` that are applied on the sources when they are fetched.

There two main reasons for this change.
- The first is that those patches (except the last one) were written in a way to improve the capability of the compiler build system to generate cross compilers, not only for Solo5, with the intention of upstreaming them. This would ease a lot the maintenance for OCaml/Solo5 (and hopefully similar projects). As we are quite advanced in the release cycle for 5.2 and so those patches would not get merged into that release, the idea is to take the time to test whether that’s really what we would require for Solo5 and propose them upstream when a strong case can be made for them.
- The second reason is that having those modifications as full patches (they are `git format-patch` patches, so they contain a full explanation message for each change) makes them easier to review. (In my review of the previous PRs, the modified Makefile of the compiler was broken so `make` was generating warnings, even if it was still able to build the compiler; using actual patches makes it easier to make sure what modified build system we will actually rely on)

One interesting outcome of this deep revision of the compiler build system is the fact that the `.opt` versions of the compilers are built, which should give better performance, in particular to build large unikernels.

Open question for those patches: I kept `git format-patch` default of using a binary diff (for `configure`, as it is declared binary in the compiler) as what is useful to review is the changes to `configure` sources that are always bundled in the same patch. `git format-patch -a` patches would be bigger (twice the size, with the current patch series) but they might be preferred for other reasons.

### A specific toolchain

The package now creates and installs a `{aarch64,x86_64}-solo5-ocaml-*` toolchain, in a similar fashion to what Solo5 does. The main reason for creating that toolchain in the first place was to avoid baking the _build-time_ directories containing `nolibc` and `openlibm` into the generated OCaml compilers. Two versions of the toolchain are generated: one with the build-time directories, which is added to `PATH` during the build of the compiler; one with the final destination directories, which is installed in `bin`.
Besides making it possible to change the `-I` and `-L` between build-time and final versions, I further used this idea to:
- store in the `cc` all its other required options (such as `-include _solo5/overrides.h`), since this machinery is rather well-placed there,
- wrap _all_ the binutils similarly; as an end-user, I find it handier to have all the prefixed tools available rather than having to remember for each of them which tool I should call (`aarch64-solo5-none...*`? `aarch64-linux-gnu-*`? ...). This fits in nicely with OCaml configure’s toolchain discovery so that stating the `--target` is enough to define all the tools of the toolchain.

### Installation

This PR proposes to delegate the actual installation to OPAM (or `opam-installer`) by creating a `ocaml-solo5.install` file rather than invoking the `install` command itself. The initial reason for this choice is that it is fairly easy to bend OCaml’s `make install` into generating the list of all the files it could install and then tell OPAM to install them only if they were actually built. So we can skip building parts of the compiler (`ocamllex`, etc.) _just_ to keep `make install` happy when they don’t make sense.

This comes with the natural consequence that it follows OPAM’s conventions when installing files, so the compiler is installed in `%{prefix}%/lib/ocaml-solo5` instead of `%{prefix}%/solo5-sysroot` (I find following the OPAM convention a rather good idea anyhow). If we relax the name of the findlib toolchain, it becomes possible to install simultaneously `ocaml-solo5` and `ocaml-solo5-cross-aarch64` (in my testing, I didn’t switch between the two as much as I would have if they had been co-installable), since the only common file between those two packages is the findlib’s `solo5.conf`.

### Smaller changes

Quite a few commits are clean-ups for things that were natural to fix while working on this (such as ensuring that `make` will rebuild `nolibc` and `openlibm` libraries iff that should be done). I feel that those small changes will make maintenance smoother.

Another change from previous PRs is the choice to drop `stdatomic.h` from `nolibc` and go with the one of the underlying compiler, to make sure the C compiler will compile C atomics as it is intended to do. This only needs a small patch to Solo5 for FreeBSD, to pull it along with the other freestanding headers when installing the Solo5 toolchain.

## Testing

This has been tested on:
- Debian x86_64 with gcc, target aarch64, building (and briefly running) the `dns-resolver` hvt mirage unikernel,
- Debian x86_64 with gcc, target x86_64, at least running the internal tests,
- FreeBSD 14 x86_64 with clang, target x86_64, running the internal tests; this needs a small patch on Solo5, to access the `stdatomic.h` header, see Solo5/solo5#574
- OpenBSD 7.3 x86_64 with clang, target x86_64, running the internal tests
- OpenBSD 7.5 x86_64 with clang: Solo5 is currently broken there, Solo5/solo5#573 has been opened there with some steps towards supporting the latest OpenBSD.

## Open question

I have been wondering about how interrupts are set up in Solo5 (without taking the time to really dig, in particular because the precise set up depends on the binding), because the way the OCaml 5 runtime sets up its stacks is really different from OCaml 4.
In particular, the OCaml 5 runtime assumes that only OCaml code should ever be triggered on a fiber stack. Is Solo5 already currently set up so that it can be guaranteed?
Commit “WIP Disable the x86_64 red zone” is strongly related to that question, on a slightly different angle.

## To do before merging

- [x] the questions in the “WIP” commits should be decided
- [x] the initial supported version of OCaml should be decided (last “WIP” commits)
- [x] `README` should be updated
- [x] dependency on `ocaml-src` should be restored
- [x] `TODO`s should be addressed
- [x] `DROPME` commits should be dropped


dinosaure:
> I have been wondering about how interrupts are set up in Solo5 (without taking the time to really dig, in particular because the precise set up depends on the binding), because the way the OCaml 5 runtime sets up its stacks is really different from OCaml 4.
In particular, the OCaml 5 runtime assumes that only OCaml code should ever be triggered on a fiber stack. Is Solo5 already currently set up so that it can be guaranteed?
Commit “WIP Disable the x86_64 red zone” is strongly related to that question, on a slightly different angle.

To contextualize the Solo5 side, I found:
- this commit which introduces `-mno-red-zone` for bindings: https://github.com/Solo5/solo5/commit/2f1cae37d8b9f97826ea08934558a4360bdb13da
- this issue (still open): https://github.com/Solo5/solo5/issues/292

shym:
Thank you for the pointers. How to properly setup interrupt handling on x86_64 is quite some rabbit hole! To sum up what I learnt:

- switching stack automatically on interrupt with no priviledge change is possible on x86_64 with IST,
- setting up the hardware correctly for a real bare-metal kernel is terribly tricky [(1)],
- but Solo5 makes a very limited and controlled use of interrupts, so that it is (hopefully) bypassing the reentrance problem,
- if there is nevertheless a reentrance problem for interrupts, the problem is happening on the interrupt stacks, there should not be problems on the stacks as seen by the OCaml runtime.

So I’ve dropped the commit disabling the red zone.
(And pinned the branch of Solo5 to fix FreeBSD support)

[(1)]: https://lore.kernel.org/lkml/20230403140605.540512-1-jiangshanlai@gmail.com/

hannesm:
Dear @shym, thanks for rebasing. Would you mind to mark it as ready for review?

shym:
I’ve rebased on `main` (cc @palainp) and added a few small fixes (left explicitly as `fixup!...` commits for now, to make it easier to resync for reviews that were already under way).
It was marked as draft (maybe wrongly, I see) as it should not be merged before OCaml 5.2 is out, but it is ready for review as is.

nolibc/mmap.c
```diff
@@ -0,0 +1,68 @@
...
+
+  void *ptr = NULL;
+  /* XXX(palainp): Does it worth to have a test on addr here? */
+  if (addr == NULL) {
+    /* Solo5 may returns -1 and set errno on error, just return MAP_FAILED.
+       It doesn't modify ptr on error: ptr will still be NULL
+     */
+    posix_memalign(&ptr, OCAML_SOLO5_PAGESIZE, len);
+  } else {
+    ptr = malloc(len);
```
shym, line 41: If I’m not mistaken, `mmap` is called with a non-`NULL` argument in the OCaml code base only in the [`map_fixed`] function, which is used only in [`allocate_minor_heap`]. But it is used there assuming (it could use at least an extra `assert`!) that the allocator will validate the _hint_ address, as it ignores the returned value except to test whether it is `MAP_FAILED` and `domain_self->minor_heap_area_start` is not updated.

[`map_fixed`]: https://github.com/ocaml/ocaml/blob/5.2.0-rc1/runtime/unix.c#L488
[`allocate_minor_heap`]: https://github.com/ocaml/ocaml/blob/5.2.0-rc1/runtime/domain.c#L498

shym (commented):
I notice only now that something does look just right in the implementation of `mmap` with a non-`NULL` argument.

nolibc/mmap.c
palainp: I understand the same. It's a bad thing because dlmalloc doesn't seem to support a hint where to allocate something :(

shym:
I dug deeper into the usage of `mmap` in the OCaml code base and got reassured: if I understand correctly, `mmap` is used with a (non-`NULL`) address only with `MAP_FIXED` to commit and decommit memory (set its access to `rw-` or `---`) that has been previously reserved. So we should be fine just ignoring those `mmap`s, since we don’t handle memory protection. :sweat_smile:
I added commit `Handle mmap(addr..., MAP_FIXED, ...) uses` to change the behaviour accordingly.
I would have been reassured about my understanding if I had managed to find an example that follows that code path, but the small programs I ran didn’t commit/decommit memory.

Finally, I also updated the PR to OCaml 5.2 RC1, released on Friday.

shym:
I’ve updated the PR to OCaml 5.2.0 as it was released last week.

Note: I’ve reverted to depend on the `ocaml` package (as it was before) rather than a precise dependency on a compiler package (`ocaml-base-compiler` or `ocaml-variants`, which was more relevant for pre-release versions of the compiler). But the build will fail later if the compiler in `PATH` is not exactly 5.2.0 because I left a strict test of version equality in the `configure` script. I don’t think it’s worth trying to relax that constraint (ie allow to build the 5.2.0 cross-compiler in a 5.2.0~rc1 switch) so I feel more confident keeping it that way.

palainp:
Thank you @shym for that update! I can confirm that this PR compiles and qubes-mirage-firewall compiles & runs as with 4.14.2 (first fast.com tests show around 10-15% more bandwidth now but this wasn't conducted with a strict evaluation protocol :) ). I'll also try with longer runs.

dinosaure:
On my side, I approve this PR. It took time for several people and I think the review of the various maintainers and authors has been done. Thank you @shym for taking over and considering what may have blocked us in our previous PRs, I consider that this one can be merged :+1:. /cc @mirage/core

palainp:
I also approve this PR, but as it breaks the supported compilers versions, and I think we also need to keep a 4.4 branch corresponding to the current head before merging :)

shym:
I’ve just updated the patches to the compiler so that it contains the patches that are already integrated upstream and in the version integrated upstream: it will make it much easier to re-synchronize when updating the patches for the next version.
I’ve also been working on the other patches, to generalize them to cover more cross-compiler use cases, but I judged it best not to update them that late in the review process and considering that they are not yet upstreamed.
If you want to have a look, it might be easier to compare the branches:
- before that update: [solo5](https://github.com/shym/ocaml/tree/solo5),
- after that update: [solo5-v1.5](https://github.com/shym/ocaml/tree/solo5-v1.5),
- the currently cooking version of the patches, if you will adventurous: [solo5-v2](https://github.com/shym/ocaml/tree/solo5-v2).

hannesm:
I merged the other PRs outstanding, Would you mind to rebase? Also, the cirrus CI should now that solo5 0.9 is released, work fine!?

shym:
I rebased on `main`. `git range-diff bd6b8dfecdd2..0b3b7ff7ae4dc8 b993ec677dcb31..99c2ea45405e82` is as expected pointing only the commits changing CI.

hannesm:
another remark: we used to depend on the ocaml-src package, now with e6caff2261d17cb32f4d7d583c275b2ddd87958a there's (instead?) and extra-spirce introduced. why? from my point of view, keeping the unnecessary changes minimal would be appreciated. if you're eager for such a change, could you propose such a pr for the main branch?

README.md
```diff
@@ -71,8 +71,7 @@ Run: `solo5-hvt _build/solo5/main.exe`
 
 ## Supported compiler versions
 
-Tested against OCaml 4.12.1 through 4.14.1. Other versions may require
-changing `configure.sh`.
+Tested against OCaml 5.0. Other versions may require changing `configure.sh`.
```
hannesm, line 74: 5.2

Makefile
```diff
@@ -1,138 +1,194 @@
...
+ocaml:
+	mkdir -p $@
+	if test -f ocaml-src.tar.gz; then \
+	  if tar --version >/dev/null 2>&1; then \
+	      tar -x -f ocaml-src.tar.gz -z -C $@ --strip-components=1; \
+	    else tar -x -f ocaml-src.tar.gz -z -C $@ -s '/^[^\/]*\///'; \
+	  fi ; \
+	elif opam var ocaml-src:lib; then cp -R `opam var ocaml-src:lib` $@; \
+	else echo Cannot find OCaml sources; false; \
+	fi
```
hannesm, line 107: this whole dance is to support sth else than ocaml-src opam package. I don't quite understand

Makefile
```diff
@@ -1,138 +1,194 @@
...
+OCAML_IS_BUILT := _build/ocaml_is_built
+$(OCAML_IS_BUILT): ocaml/Makefile.config | _build
+	PATH="$(abspath $(TOOLDIR_FOR_BUILD)):$$PATH" $(MAKE) -C ocaml cross.opt
+	cd ocaml && ocamlrun tools/stripdebug ocamlc ocamlc.tmp
+	cd ocaml && ocamlrun tools/stripdebug ocamlopt ocamlopt.tmp
+	touch $@
+
+# TODO: Decide whether these files should be provided in the repository (if so,
+# we should make sure they cover all the use cases and move the files to the
+# root) or rebuilt every time
```
hannesm, line 137: would be great to solve this TODO before this is being merged.

Makefile
```diff
@@ -1,138 +1,194 @@
...
-	$(RM) -r ocaml/
-	$(RM) solo5.conf
+	$(RM) -rf _build
 	$(MAKE) -C openlibm clean
-	$(MAKE) -C nolibc \
-	    "FREESTANDING_CFLAGS=$(NOLIBC_CFLAGS)" \
-	    "SYSDEP_OBJS=$(MAKECONF_NOLIBC_SYSDEP_OBJS)" \
-	    clean
+	$(MAKE) -C nolibc clean FREESTANDING_CFLAGS=_
+	if [ -d ocaml ] ; then $(MAKE) -C ocaml clean ; fi
```
hannesm, line 175: again, it is not very clear to me why to handle the case that `ocaml` directory does not exist.

Makefile
```diff
@@ -1,138 +1,194 @@
...
+	$(RM) -f $(INSTALL_FILES)
 
+.PHONY: distclean
 distclean: clean
-	rm Makeconf
+	$(RM) -f Makeconf
+# Don't remove the ocaml directory itself, to play nicer with
+# development in there
+	if [ -d ocaml ] ; then $(MAKE) -C ocaml distclean ; fi
```
hannesm, line 183: same as above, unclear why this conditional is here.

nolibc/Makefile
```diff
@@ -42,15 +43,15 @@ TEST_H_OBJS=$(patsubst %.h,test-%.o,$(HEADERS))
...
 
-test-include:
-	mkdir $@
+test-include/:
+	mkdir -p $@
 
-test-include/sys: test-include
-	mkdir $@
+test-include/sys/: test-include/
+	mkdir -p $@
```
hannesm, line 55: these changes, adding a trailing `/`, are they worth it? if yes, should they be PRed to the main branch?

nolibc/include/errno.h
```diff
@@ -2,13 +2,18 @@
...
+#define ENOMEM      12    /* Out of memory */
+#define EBUSY       16    /* Device or resource busy */
+#define EINVAL      22    /* Invalid argument */
+#define EMFILE      24    /* Too many open files */
+#define ERANGE      34    /* Math result not representable */
+#define ENOSYS      38    /* Invalid system call number */
+#define EOVERFLOW   75    /* Value too large for defined data type */
+#define EINTR        4    /* Interrupted system call */
+#define EPIPE       32    /* Broken pipe */
+#define ECONNRESET 104    /* Connection reset by peer */
```
hannesm, line 17: why this renumbering?

nolibc/include/fcntl.h
```diff
@@ -3,11 +3,12 @@
...
-#define O_CREAT (1<<3)
-#define O_TRUNC (1<<4)
-#define O_EXCL (1<<5)
+#define O_RDONLY 00000000
+#define O_WRONLY 00000001
+#define O_RDWR   00000002
+#define O_CREAT  00000100
+#define O_EXCL   00000200
+#define O_TRUNC  00001000
+#define O_APPEND 00002000
```
hannesm, line 12: what is the reasoning behind this change?

nolibc/include/signal.h
```diff
@@ -1,17 +1,27 @@
...
-#define SIG_ERR NULL
-/*
- * The following definitions are not required by the OCaml runtime, but are
- * needed to build the freestanding version of GMP used by Mirage.
- */
+#define SIG_DFL 0
+#define SIG_IGN 0
+#define SIG_ERR 0
+#define SIG_BLOCK 0
+#define SIG_SETMASK 0
```
hannesm, line 18: NULL vs 0 - what is the reason to use 0 now?

nolibc/include/string.h
```diff
@@ -10,17 +10,24 @@ void *memmove(void *, const void *, size_t);
...
+char *strcpy(char *, const char *);
 char *strchr(const char *, int);
 char *strstr(const char *, const char *);
 /*
  * The following definitions are required for the OCaml bytecode runtime.
  */
 int strncmp(const char*, const char*, size_t);
 
+/* TODO(dinosaure): we must track where they are used to know if we need
+ * to implement them or not. */
```
hannesm, line 29: this TODO, what is the state with experimentation? can the TODO be removed? can the function prototypes be discarded?

hannesm:
Thanks for your work. Above I added some remarks. I have some questions:
- the ocaml-src dependency, is dropping it worth it? If yes, I'm still confused by the `if [ -d ocaml ]` in the Makefile, so why would there be support for no ocaml directory?
  - If it is worth dropping, and using the extra-sources, please document the reasoning behind it.
- which OCaml releases should this target? 5.2.0? 5.2.1? or directly 5.3.0?

In your initial message, you write:

> To do before merging
>
> - “TMP” commits should be updated to OCaml 5.2
> - the questions in the “WIP” commits should be decided
> - README should be updated

Has this been addressed?

Now, you introduce some shell scripts (as I understand it to remove sed from the Makefile). I'd appreciate if shellcheck.net would be used on each shell script to have a nicely behaving script. I pasted the shellcheck output below.

configure.sh:
```bash
[Line 3:](javascript:setPosition(3, 23))
prog_NAME="$(basename $0)"
                      ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.

Did you mean: ([apply this](javascript:applyFixIndex([0])), apply [all SC2086](javascript:applyFixCode(2086)))
prog_NAME="$(basename "$0")"
 
[Line 7:](javascript:setPosition(7, 32))
    echo "${prog_NAME}: ERROR: $@" 1>&2
                               ^-- [SC2145](https://www.shellcheck.net/wiki/SC2145) (error): Argument mixes string and array. Use * or separate argument.
 
[Line 12:](javascript:setPosition(12, 32))
    echo "${prog_NAME}: ERROR: $@" 1>&2
                               ^-- [SC2145](https://www.shellcheck.net/wiki/SC2145) (error): Argument mixes string and array. Use * or separate argument.
 
[Line 76:](javascript:setPosition(76, 19))
TARGET_TRIPLET="$($CONFIG_TARGET-cc -dumpmachine)"
                  ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.

Did you mean: ([apply this](javascript:applyFixIndex([3])), apply [all SC2086](javascript:applyFixCode(2086)))
TARGET_TRIPLET="$("$CONFIG_TARGET"-cc -dumpmachine)"
```

gen_dot_install.sh
```bash
[Line 3:](javascript:setPosition(3, 20))
if [ ! -e "$1.lib" -o ! -e "$1.libexec" ]; then
                   ^-- [SC2166](https://www.shellcheck.net/wiki/SC2166) (warning): Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.
```

gen_ocaml_install.sh -- this uses `/usr/bin/env bash` -- could we stick to `/bin/sh` features?
```bash
[Line 8:](javascript:setPosition(8, 25))
  if [ ! -d "$OCAMLDIR" -o -z "$PREFIX" ]; then
                        ^-- [SC2166](https://www.shellcheck.net/wiki/SC2166) (warning): Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.
```

gen_solo5_conf.sh: congratulations, no warnings

gen_toolchain_tool.sh
```bash
[Line 28:](javascript:setPosition(28, 6))
  if $SOLO5_TOOLCHAIN-cc -dM -E - </dev/null | grep -Eq '^#define __clang__ 1$'
     ^-- [SC2086](https://www.shellcheck.net/wiki/SC2086) (info): Double quote to prevent globbing and word splitting.

Did you mean: ([apply this](javascript:applyFixIndex([0])), apply [all SC2086](javascript:applyFixCode(2086)))
  if "$SOLO5_TOOLCHAIN"-cc -dM -E - </dev/null | grep -Eq '^#define __clang__ 1$'
```

hannesm:
The reason why I think that the "extra-sources" is not the best approach is in previous versions / releases, we supported multiple OCaml versions with the same ocaml-solo5 release. Now, this only works due to the dependency on `ocaml-src`.

For this PR and the future, I'm happy to hear what kind of support story you have in mind? Since OCaml 5 is still rather experimental, I'm fine to say "let's target the latest release only".

nolibc/include/signal.h
palainp: Now sure about that, according to me, NULL has a pointer semantic and signal are defined with integers. As solo5 does not use signals, there is no risks except maybe some conversion warning?

palainp (commented):
Thank you so much @hannesm for your review, I tried to provide an answer for some of your questions :)

shym:
Thank you for all the feedback and sorry for the latency.

> I think you can remove the pin now, and depend on >= 0.9.0...

Indeed! Done.
(I start going through all the other comments)

nolibc/include/errno.h
hannesm: great, would you mind to reorder and provide a comment where these values are coming from?

nolibc/include/fcntl.h
hannesm: great, would you mind to add a comment where these values in this form are coming from? maybe your linux distribution with a version indicator?

nolibc/include/signal.h
hannesm: 0 is fine, thanks for your reply.

shym:
> then the question is: 5.2.1 support? 5.3? what should we target with an initial merge and release?

Maybe we could indeed skip 5.2.0 now that 5.2.1 is released. I’ll port my patches.
As all the test we have run so far are for 5.2, I think it’s worthwhile to release a version that’s compatible with 5.2.x, if only to compare the performance with 5.3. We can always deprecate the 5.2.x later if we want to drop support for it?

> we used to depend on the ocaml-src package […]

I must confess I’ve now forgotten why I went with this. I _suppose_ the reason was exactly the situation we are in just now: OCaml 5.2.1 was released but the `ocaml-src` package is not updated yet, and I wanted this PR to be easy to test out (without creating an overlay for `ocaml-src`, etc.). Now, I agree that relying on `ocaml-src` is a better long-term solution. I’ll do that.

> I'm still confused by the `if [ -d ocaml ]` in the Makefile, so why would there be support for no ocaml directory?

It is somehow the equivalent of `-f` in the `rm -f` used in `clean` instructions: the rationale is that `make clean` should not fail if `ocaml` has not been created yet, whatever the reason. It might be when the user explicitly removed it to make sure it will be rebuilt from scratch, as the Makefile explicitly doesn’t do so even on `distclean`, which I find a much more convenient behaviour for developing `ocaml-solo5`.

> In your initial message, you write:
>
> > To do before merging
> […]
>
> Has this been addressed?

Not yet. In particular as those either:

- depend on the version(s) of the compiler we target for this PR (for TMP and README commits),
- were open questions for review comments (for WIP commits).

I’ll assume that I can move and clean up WIP commits as I see best now, I guess.

[To be continued :-)]

shym:
> relying on `ocaml-src` is a better long-term solution.

Digging around (in particular mirage/mirage-platform#157), I see that two solutions seem to coexist to find `ocaml-src`: the `opam var ocaml-src:lib` I’ve used or `ocamlfind query ocaml-src`. Any opinion about the best choice to do here?
(I wonder whether `ocamlfind query` might have a better chance of playing well with `dune pkg`, maybe? It also seems to be consistently faster in my tests locally).

hannesm:
> Digging around (in particular [mirage/mirage-platform#157](https://github.com/mirage/mirage-platform/pull/157)), I see that two solutions seem to coexist to find `ocaml-src`: the `opam var ocaml-src:lib` I’ve used or `ocamlfind query ocaml-src`. Any opinion about the best choice to do here? (I wonder whether `ocamlfind query` might have a better chance of playing well with `dune pkg`, maybe? It also seems to be consistently faster in my tests locally).

as you like, bonus points to keep the main branch and this PR in sync in these regards (and I'm fine to have the main branch -- soon we should fork to a ocaml-4 branch being changed)

dinosaure:
> Any opinion about the best choice to do here?

nolibc/Makefile
shym: This was not really to me either but as it doesn’t hurt, I kept commit a9a3ffd23cc5c9177b78170698c10e987d9c22f0 as is.

Makefile
shym: I answered in https://github.com/mirage/ocaml-solo5/pull/134#issuecomment-2490916587.

Makefile
shym: Hopefully answered in https://github.com/mirage/ocaml-solo5/pull/134#issuecomment-2490916587.

shym:
Rebased after the merge of #141 to fix the shellcheck-reported errors.

nolibc/include/errno.h
palainp: Done in https://github.com/shym/ocaml-solo5/pull/1. I prefered to PR against the branch instead of directly push commits to be able to discuss there if needed. There was some mistakes in the numbers, I've updated them in the PR with the actual values.

nolibc/include/fcntl.h
palainp: Done in https://github.com/shym/ocaml-solo5/pull/1 and for consistency I also used the numbers in the BSD libc as for errno.h.

shym:
I updated the branch so that it relies on `ocaml-src` to get the compiler sources and `ocamlfind` to find those sources.
It also now provides an implementation of `strerror_r` as it is actually used by the runtime. So I’ve reverted to using sequential integers for error codes, so that the implementation of `strerror_r` can simply pick the error message in an array with no useless empty cells. The error messages are placed in a separate file, borrowed from OpenBSD but keeping only the entries we use.

I’ve also added a small (silly) example unikernel that raises an exception using an `strerror_r`-generated payload, if only to make sure that it is indeed used. As expected, without the implementation of `strerror_r`, the [unikernel terminated with](https://github.com/shym/ocaml-solo5/actions/runs/12016202136/job/33495895119#step:9:25):
```
STUB: abort: strerror_r() called
```
instead of:
```
Fatal error: exception Sys_error("/tmp: Function not implemented")
```
(which admittedly reads funny).

nolibc/include/errno.h
```diff
@@ -3,21 +3,24 @@
...
+#define EINVAL       6
+#define EMFILE       7
+#define EPIPE        8
+#define ERANGE       9
+#define EAGAIN      10
+#define ECONNRESET  11
+#define ENOSYS      12
+#define EOVERFLOW   13
+
+#define NB_ERRORS   14
```
palainp, line 25: Maybe it is possible to use an enumeration here?
```C
enum {
    ENOERROR=0, /* Actual error codes should be > 0 */
    ENOENT,
    EINTR,
    ...
    NB_ERRORS
};
```
This permits to keep the value for `NB_ERRORS` up to date and to avoid any mangling with numbers. WDYT?

nolibc/strerror_r.c
```diff
@@ -0,0 +1,19 @@
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+
+extern const char *const sys_errlist[];
+
+int strerror_r(int num, char *buf, size_t buflen)
+{
+	if (num < 0 || num >= NB_ERRORS) {
```
palainp, line 9: This will work as we have a string for `num==0`, but as 0 is not a valid error number, maybe it's better to have `(num <= 0 || num >= NB_ERRORS)`?

shym: glibc considers 0 as a valid argument for `strerror_r`, so I’d rather follow their choice there.

nolibc/include/errno.h
shym: My immediate reaction is: I don’t think “people” do that. By which I mean I usually see lists of `#define`, in OpenBSD sources for instance. I have no real reason to pick one or the other in that case though.

nolibc/strerror_r.c
palainp: And BTW the string is "Success", which sounds good for that value of `errno` :)

nolibc/include/errno.h
shym: My feeling is that what will be painful to update will be the error-message list and keeping the two lists in sync, not so much keeping the values sane. It might be because I know my `vim` pretty well :smile:

shym:
I updated the PR with 5.2.1 support (with a temporary overlay for `ocaml-src`). I plan to drop 5.2.0 when `ocaml-src.5.2.1` is available in your nearest opam repo.

nolibc/Makefile
hannesm: ok, should we single that commit out and push to the main branch?

hannesm:
thank you, so the ocaml-src is now part of the opam-repository. and sorry, there are now merge conflicts since the main branch has been updated.. would you mind to rebase this branch?

is this then fine to merge? as far as I can tell, the reviews are addressed (apart from the minor https://github.com/mirage/ocaml-solo5/pull/134#discussion_r1847343254)?

hannesm:
thanks. looks fine to me. should we merge and release?

palainp:
From my POV yes :) 👍 Maybe we can also create a branch for the 4.14 version ?

dinosaure:
> From my POV yes :) 👍 Maybe we can also create a branch for the 4.14 version ?

Merged, thanks for all you. It was a really collaborative work, which is always very satisfying :tada:. I made a branch as you asked @palainp. Thanks again, it was a huge work!

PR #135, opened 2024-04-23
fix memory usage for posix_memalign

hannesm:
Dear @palainp, I prepared here a PR of your branch to track and review it properly. :) Thanks for your investigation and work.

Indeed, in the solo5-hvt/spt/virtio setting, the posix_memalign is never called as far as I can tell -- but xen requires it (due to io-page).

### DIFF
--- a/nolibc/dlmalloc.i
+++ b/nolibc/dlmalloc.i
@@ -5,7 +5,8 @@
   memory allocated in the heap.
   The results of this fast estimation are returned by the
   dlmalloc_memory_usage function exported as malloc_memory_usage.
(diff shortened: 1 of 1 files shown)


hannesm:
to me, this looks fine. happy to see further reviews. @palainp also please indicate whether you're happy with your commit :)

palainp:
Thank you @hannesm , this change is ok to me with qubes-mirage-firewall after some hours of running (I saw free memory going up and down as expected).

hannesm:
thank you for your work, I cut the 0.8.4 release with your fix https://github.com/ocaml/opam-repository/pull/25717

palainp:
Thanks! I'll update qubes-mirage-firewall with that realease :)

PR #136, opened 2024-05-06
Make the mode appear clearly in CI

shym:
This PR is a small quality-of-life improvement, by making the mode standout in CI results, as I have been repeatedly looking for the `spt` mode to make sure the test ran without any surprising log (I suspect the logs of the other modes are not much more informative than just their success result, when successful).


hannesm:
thanks!

PR #137, opened 2024-05-31
Clean up the test and its use in CI

shym:
This PR combines a couple of small details that have been nagging me in the test and, partly as a consequence, also in CI.
The main changes are:

- import a fix to the test programs by Romain and Pierre extracted from #124/#134,
- deduplicate `example` and `test` as they are really the same code (I kept the name `example` to emphasize its use as documentation, but with the multiple backends of the `test`, obviously),
- simplify its dune rules, so that the possibility to test multiple backends is noiseless,
- test all the backends in the same job as I noticed that in CI runs all the time is spent in the setup, not in the test itself, so it is wasteful not to share one setup to run the tests of all the backends.

I hesitated to break it down into smaller PRs but those changes are fairly tightly coupled so the smaller PRs would have conflicted quite a bit. But that means I could easily drop some of changes this PR proposes.


hannesm:
I like the changes, and when CI passes would be happy to merge them! Thanks a lot.

dinosaure (approved):

PR #138, opened 2024-07-21
Update GitHub Actions

smorimoto:
https://github.com/smorimoto/ocaml-solo5/actions/runs/10027310209


hannesm:
Thanks for your PR, I rebased and force-pushed on top of #140 to hopefully see green checkmarks here all over :D

PR #142, opened 2024-11-22
use ocamlfind instead of opam to discover where ocaml-src is located

hannesm:
and where to install and uninstall to

Discussed in https://github.com/mirage/ocaml-solo5/pull/134#issuecomment-2491605469 and https://github.com/mirage/ocaml-solo5/pull/134#issuecomment-2491649732 by @shym and @dinosaure


shym:
I wonder if using `$(ocamlfind query ocaml-src)` to find parent directories is not a bit fragile: if this is not in an OPAM-like setup, we could end up installing and removing stuff in wrong directories. I would be more comfortable for those scripts to error out if the prefix is not explicitly set.

dinosaure:
May be `ocamlfind printconf destdir` is preferable. However, `prefix` seems highly related to the OPAM layout.

hannesm:
thanks for your feedback, I removed uninstall (it is unused anyways), and adapted install.sh to use `$PREFIX`.

dinosaure (approved):

hannesm:
the remaining question I have is how defensive should `install.sh` be? should we warn and fail on all potential issues (PREFIX being undefined, MAKE being undefined, ...)? Or is it fine to just use the empty prefix, an undefined MAKE, ...? Since it is called by `make install` (and there's no expectation to call it differently), I'm fine with the no error handling approach as proposed in this PR.

If there's some strong opinion to act differently, please raise your voice - best with a brief explanation why it should be more defensive.

shym (approved):
This looks good to me. I agree with the approach: `install.sh` doesn’t need to accomodate for direct invocation and so `$PREFIX` should be empty only when the user has explicitly set it so in `configure.sh`, so there should be no surprise there.

hannesm:
Thanks, squash-merged.

PR #145, opened 2024-12-11
Update the README

shym:
This PR brings a couple of updates to the README. In particular it lists the components as that part of the documentation is out-of-sync after #134 (sorry about that).

It is split in 2 commits, as the first commit contains modifications that are also relevant for the `4.14` branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,14 @@
 # ocaml-solo5 -- OCaml compiler with Solo5 backend
 
 This package provides a OCaml compiler suitable for linking with a
-Solo5 base layer.
+Solo5 base layer:
+
+- package versions ≥ 1.0 and the `main` branch are compatible with OCaml 5+
+  compilers, see the “[Supported compiler versions]” section for details,
+- package versions 0.8.x and the `4.14` branch are compatible with OCaml 4.14
+  compilers.
+
+[Supported compiler versions]: #supported-compiler-versions
 
 ## License and contributions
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,11 +46,12 @@ In `PREFIX/solo5-sysroot/include/nolibc`:
 In `PREFIX/lib/findlib.conf.d`:
 
 - `solo5.conf`: ocamlfind definition of the cross-compilation switch.
+- `solo5.conf`: ocamlfind definition of the cross-compilation toolchain.
```
hannesm, line 57: within this commit, please remove the above line - so we only have `solo5.conf` once.

hannesm:
for the README.md on the main branch, can we as well describe that it targets OCaml 5 (with the 1.x releases) -- and we have the branch "4.14" for the OCaml 4.14 series (which is well be maintained, with 0.8.x releases)

hannesm:
I forgot to say: thanks for your PR :)

README.md
shym: Sorry about that and thanks. Fixed.

shym:
> for the README.md on the main branch, can we as well describe that it targets OCaml 5 (with the 1.x releases) -- and we have the branch "4.14" for the OCaml 4.14 series (which is well be maintained, with 0.8.x releases)

Good idea. I propose that in the last commit. I think it makes more sense to add that information near the beginning of the README and so I wondered whether I should pull the small “Supported compilers” section there also.

PR #146, opened 2025-01-08
add x-maintenance-intent field to the opam file

palainp:


palainp:
Thank you ! That's easier :)

PR #148, opened 2025-01-23
Support OCaml 5.3.0

shym:
This PR replaces the supported OCaml version with 5.3.0. This PR sits at the moment on top of #147 in order to illustrate with the new examples at least some of new possibilities of that version.

Most of the patches that are required for the compiler have been accepted upstream (so to be part of 5.4) and are therefore backported here. As the patches that we have in `main` for 5.2.1 are slightly incompatible (they were drafts for the patches that got upstreamed), they would need to be reworked for our main `Makefile` to support both versions. As 5.2.1 was still considered experimental, this PR proposes to simply support 5.3.0.

The main part of this PR then is the set of patches for 5.3.0. Those patches are extracted from upstream PRs (selecting the commits that are relevant for OCaml/Solo5):

- PR 13526 (merged upstream) (10 commits)
- PR 13674 (merged upstream) (1 commit)
- PR 13719 (merged upstream) (1 commit)
- PR 13735 (under review) (3 commits)
- a not-opened-yet PR to support freestanding targets (2 commits)
- a not-to-be-upstreamed commit to fix the maximum number of domains to 1

Note that since upstream PR 13272, the maximum number of domains can be set using a parameter in `OCAMLRUNPARAM` so I tried to replace that last patch by adding to `nolibc` a `secure_getenv` returning `"d=1"` for the `OCAMLRUNPARAM` value. But that value is then parsed by OCaml using `sscanf`, which is not provided in `nolibc`. I settled to keep that simple patch, at least for now, rather than come up with a `sscanf` (do we want to provide a hard-coded version that would work only for that single use case or a real implementation?).

## Commit overview

- Support for 5.3.0
  - 1d014d2 Add patches for 5.3.0
  - a6eb56a Use OCaml’s `crossopt` and `installcross` rules
  - 7a99ebc Add stubs for `qsort` and `longjmp`
- Fix our `solo5.conf` to support OCaml-provided libraries
  - 85bdebe Add OCaml’s directory to the toolchain-configuration `path`
- Update the version in opam files, CI, README
  - 76fcd25 opam: Update dependency to OCaml 5.3.0
  - e5dc6fe CI: Update OCaml version to 5.3.0
  - 911b05d CI: Show the configuration of the Solo5 toolchain
  - 641652f README: Update the supported OCaml version to 5.3.0
- Drop patches for 5.2.1 (broken by the commit updating our `Makefile`)
  - 83d9135 Remove patches for 5.2.1
- Temporary fix for the CI
  - a2dbf0f TMP CI: Pin a test branch for Solo5 on FreeBSD


shym:
I updated this PR with the latest, merged, version of an upstream PR. In particular, it means that `.size` and `.type` directives should be properly emitted, so they can be depended upon for debugging.

patches/5.3.0/0008-Add-a-Config-entry-for-the-target-OS-type.patch
```diff
@@ -0,0 +1,196 @@
+From ddf39e7e09d03f5e83c92b6bb1bc094e8f116a66 Mon Sep 17 00:00:00 2001
```
dinosaure, line 1: A very quick review, is this patch really necessary? In particular, it changes the CRCs of the stdlib, which makes it impossible for a third-party library (compiled with the host compiler) to be linked.

dinosaure: I removed this patch to avoid the boostrap and retested a simple unikernel and it seems to me that this patch is not strictly necessary to ensure that the runtime can work with Solo5. Furthermore, the CRCs of the stdlib without this patch are the same as those of the host compiler, which makes it possible to link third-party libraries compiled with the host compiler with the stdlib of ocaml-solo5.

shym: This is a good example of why I based this PR on top of #147: without that patch, the `config` example is broken, as it reports
```
OS: None
Unix: true
```
(https://github.com/shym/ocaml-solo5/actions/runs/13496596181/job/37705178517#step:7:57).
On the other hand, maybe that’s a breakage we’d rather ignore for now (as it is broken on builds for OCaml 4, IIUC, so probably not expected to be correct in Mirage unikernels).

dinosaure: > On the other hand, maybe that’s a breakage we’d rather ignore for now (as it is broken on builds for OCaml 4, IIUC, so probably not expected to be correct in Mirage unikernels).

It depends on the objectives we have, but I think a conservative approach to OCaml's stdlib would be best. In addition, I think ocaml-solo5 should focus primarily on `libasmrun.a` (and not `stdlib.cmxa`).

As for Mirage, a differentiation (whether it is ocaml-solo5 or ocaml) via `Sys.something` is not used in the Mirage ecosystem. The only (important) differentiation is found mainly in the C stubs and during compilation (and it is currently managed by `dune`) because it is really in these C files that ocaml-solo5 produces different object files.

More generally, we know when we want to obtain a unikernel in advance — specifically before building it. Having the possibility of knowing dynamically whether or not we are in a unikernel can open the door to a whole host of possibilities that we might regret.

palainp:
Dear all, thank you @shym for your work on that update :)
I currently have an issue compiling this PR with qubes-mirage-firewall:
```
$ mirage configure -t xen && make depend && dune build
File "duniverse/mirage-xen/lib/bindings/dune", lines 1-10, characters 0-162:
 1 | (rule
 2 |  (enabled_if
 3 |   (= %{context_name} solo5))
 4 |  (deps
 5 |   (source_tree .))
 6 |  (targets libmirage-xen_bindings.a)
 7 |  (action
 8 |   (no-infer
 9 |    (progn
10 |     (run %{make})))))
Context: solo5
ocamlfind -toolchain solo5 ocamlopt -ccopt "-I ./include/ -O2 -std=c99 -Wall -D__XEN_INTERFACE_VERSION__=__XEN_LATEST_INTERFACE_VERSION__"   -c -o bmap.o bmap.c
ocamlfind: Package `threads' not found
gmake: *** [<builtin>: bmap.o] Error 2
```
I don't know what I did wrong, I currently have pins on:
- the future version of mirage (4.9.0 without functors for time, mclock, pclock, which compiles and works fine with ocaml 4.14.2, so I guess it should work here too, and the error seems unrelated to the unikernel compilation),
- this PR,
- the opam-monorepo PR proposed to work with the new Ocaml 5.3 compiler (updated to allow opam 2.3).
The issue seems to be mirage-xen compilation related and I'm not familiar with ocamlfind to check what is expected of my code :)

gen_dot_install.sh
```diff
@@ -42,23 +33,18 @@ main() {
...
   printf '%s: [\n' lib
-  cat "$prefix_for_ocaml_chunks".lib
   walk_tree nolibc/include include
   install_file nolibc/libnolibc.a lib/libnolibc.a
   walk_tree openlibm/include include
   walk_tree openlibm/src include .h
   install_file openlibm/libopenlibm.a lib/libopenlibm.a
   # dummy packages
-  for pkg in nolibc threads is_solo5; do
+  for pkg in nolibc is_solo5; do
```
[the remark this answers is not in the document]
palainp: I think you're right :) but the commit message indicates that this removal is needed "Remove the, now useless, dummy file that was created to ensure `threads` could be found". So do we need to add something (e.g. the empty opam files) in mirage-xen?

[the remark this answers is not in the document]
palainp: I can confirm that reverting this part of the commit solves my error. So I'm wondering how else can I get threads?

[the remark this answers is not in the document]
palainp: Resolved with [85bdebe](https://github.com/mirage/ocaml-solo5/pull/148/commits/85bdebecdd59aec3a1427a0f36e7369110ffab4e), thank you @shym !

shym:
@palainp: thank you for testing that PR! I don’t really know why the problem was hiding on my machine. Anyway, I’ve reverted that change and added a small test in CI to make sure that `ocamlfind` accepts to run.
@dinosaure: I’ve also dropped the `Config.target_os_type` patch from the series, it should be fixed on the next OCaml release anyway.
The CI failure on FreeBSD is unrelated, see #149.

hannesm:
Before we merge and release this, we need a release of solo5 with https://github.com/Solo5/solo5/pull/590 included.

dinosaure:
Hmmhmm, I currently have this error on my machine:
```
#=== ERROR while compiling ocaml-solo5.1.0.1 ==================================#
# context     2.3.0 | linux/x86_64 | ocaml-base-compiler.5.3.0 | pinned(git+file:///home/dinosaure/dev/ocaml-solo5#530#a2dbf0f05e0319a017d6df5dd5ffc551eedc9e47)
# path        ~/.opam/5.3.0/.opam-switch/build/ocaml-solo5.1.0.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build make -j31
# exit-code   2
# env-file    ~/.opam/log/ocaml-solo5-126344-db3a24.env
# output-file ~/.opam/log/ocaml-solo5-126344-db3a24.out
### output ###
# Error: The files "/home/dinosaure/.opam/5.3.0/.opam-switch/build/ocaml-solo5.1.0.1/ocaml/stdlib/stdlib.cmi"
# [...]
#        make inconsistent assumptions over interface "Stdlib"
# make[2]: *** [Makefile:2541: lambda/switch.cmx] Error 2
# File "/home/dinosaure/.opam/5.3.0/.opam-switch/build/ocaml-solo5.1.0.1/ocaml/parsing/camlinternalMenhirLib.ml", line 1:
# Error: The files "/home/dinosaure/.opam/5.3.0/.opam-switch/build/ocaml-solo5.1.0.1/ocaml/stdlib/stdlib.cmi"
#        and "/home/dinosaure/.opam/5.3.0/.opam-switch/build/ocaml-solo5.1.0.1/ocaml/parsing/camlinternalMenhirLib.cmi"
#        make inconsistent assumptions over interface "Stdlib"
# make[2]: *** [Makefile:2541: parsing/camlinternalMenhirLib.cmx] Error 2
# make[2]: Leaving directory '/home/dinosaure/.opam/5.3.0/.opam-switch/build/ocaml-solo5.1.0.1/ocaml'
# make[1]: *** [Makefile.cross:84: crossopt] Error 2
# make[1]: Leaving directory '/home/dinosaure/.opam/5.3.0/.opam-switch/build/ocaml-solo5.1.0.1/ocaml'
# make: *** [Makefile:124: _build/ocaml_is_built] Error 2
```

I can not understand it now but I will try to figure out where is the error (note that I was able to delete the boostrap patch on myself and compile `ocaml-solo5`).

dinosaure:
I looked at this error for a while and it seems to be related to my installation of OCaml by Archlinux. It seems that the version for Archlinux has been “slightly” modified. The checksum of stdlib.cmi is therefore different from the one produced by ocaml-solo5, hence the above error. So I uninstalled OCaml (via pacman) and reinstalled OPAM as well as a new switch OCaml 5.3.0 whose sources are indeed the same as the one used by ocaml-solo5 and I no longer get the error. Not sure if it's related to this PR but the error is still strange.

Despite this point, I agree with this PR.

palainp (approved):
I'm happy with this PR, however I've tested only on linux, so it's important to wait for https://github.com/Solo5/solo5/pull/590 to be merged and solo5 released for CI to be happy with freeBSD.
Should we also wait for https://github.com/ocaml/ocamlfind/pull/99 to be merged and released (and `threads` removed again)?

shym:
> Should we also wait for [ocaml/ocamlfind#99](https://github.com/ocaml/ocamlfind/pull/99) to be merged and released (and `threads` removed again)?

I would not say so, we can keep using the same workaround we’ve been using for a long time, and drop it if/when upstream is fixed.

shym:
I rebased it, but it is still waiting for a release of Solo5 to support FreeBSD, so that the TMP commit can be dropped.

palainp:
Now solo5 0.9.1 is released and merged in opam-repository, I think I can drop the last commit to trigger a CI test, is that ok for you @shym ?

shym:
Do you understand the CI error on FreeBSD?
(Feel free to fix it if you do)

hannesm:
> Do you understand the CI error on FreeBSD? (Feel free to fix it if you do)

to me it looks like there's still somwtiing pinning solo5 to your branch. maybe the caching is bad?

```
(git+file:///tmp/cirrus-ci-build#HEAD)
[ERROR] Could not synchronize /.opam/5.3.0/.opam-switch/sources/solo5 from "git+https://github.com/shym/solo5.git#freebsd-stdalign":
        Branch freebsd-stdalign not found
```

palainp:
I agree the issue is likely due to the pin still present, I tried to trigger CI again, but I can see the pin is still present :(

hannesm:
My proposal: remove the cache stuff in .cirrus.yml. I don't know much about that caching (not sure who proposed it) -- but better have a correct CI than one with a bad caching ;) -- alternatively debug the caching...

palainp:
So I made things worse by playing with the cache :( I'm going to look for and read some docs :)

palainp:
It seems to be not related with me fiddling with cache (https://github.com/ocaml/setup-ocaml/issues/943). So I'll re-remove the cache part later when that issue will be solved.

palainp:
\o/ CI takes longer but is green now :)

dinosaure:
> This is really weird.
> I wonder:
> 
> - what is in the difference between this PR and your branch where it worked?
> - whether the Arch package is reproducible?

It seems that the Archlinux version compile with `-ffat-lto-objects`. I suspect that it changes in a way (not sure why) `*.cmi` outputs. `ocaml-solo5` expects, in some way (as far I understand), that the `stdlib` provided by the host system should be exactly the same as the one it tries to compile. I think that's the root of the issue.

palainp:
> > This is really weird.
> > I wonder:
> > 
> > * what is in the difference between this PR and your branch where it worked?
> > * whether the Arch package is reproducible?
> 
> It seems that the Archlinux version compile with `-ffat-lto-objects`. I suspect that it changes in a way (not sure why) `*.cmi` outputs. `ocaml-solo5` expects, in some way (as far I understand), that the `stdlib` provided by the host system should be exactly the same as the one it tries to compile. I think that's the root of the issue.

Oh sorry, I merged to fast, I thought it was OK. Do you want to fix that in a new PR and merge before a release?

dinosaure:
> Oh sorry, I merged to fast, I thought it was OK. Do you want to fix that in a new PR and merge before a release?

It's ok but I think it arises the question of the reproducibility expected by `ocaml-solo5` when it comes to load `*.cmi` from the host system.

PR #149, opened 2025-02-26
CI: Update the FreeBSD version

shym:
The FreeBSD 14 image provided on the CI machines has changed to 14.2. I didn’t find a way to ask simply for the latest version though.


hannesm:
Merci beaucoup

PR #154, opened 2025-09-26
Add an `opatch` command and use it instead of `git apply`

shym:
This PR is a draft, in the sense that it’s probably not meant to be merged as is, as it is more of a prototype than a finished version. But it would address #151.

This PR adds an `opatch` command that uses the `patch` library and is largely extracted from opam (so kudos to the original authors there!). I think it is of more general use than just `ocaml-solo5` (at least `ocaml-unikraft` is of interest here :-)) so it might be nice to push it into another repository, maybe `patch` itself if the command is good enough. I’ve cut a few corners as not everything was actually useful for our use case (in particular to delete directories that end up empty after patching).

As half of `opatch` is extracted from opam, it inherits its LGPL license; if we want to relicense it under ISC like `patch`, we would need the original authors’ agreement (`apply` is mostly @kit-ty-kate’s code, if I’m not mistaken; `realpath` has more history) or rewrite them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,17 +90,23 @@ $(TOOLDIR_FINAL)/$(MAKECONF_TARGET_ARCH)-solo5-ocaml-%: \
 .PHONY: toolchains
 toolchains: $(TOOLCHAIN_FOR_BUILD) $(TOOLCHAIN_FINAL)
 
+# OPATCH
+opatch: opatch.ml
+	ocamlfind opt -package unix -package patch -linkpkg $< -o $@
+
 # OCAML
 # Extract sources from the ocaml-src package and apply patches if there any in
 # `patches/<OCaml version>/`
-ocaml:
+ocaml: | opatch
 # First make sure the ocaml directory doesn't exist, otherwise the cp would
 # create an ocaml-src subdirectory
 	test ! -d $@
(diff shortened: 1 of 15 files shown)


kit-ty-kate:
Concerning the license, `real_path` is the only function taken from opam that's not my code and it is not needed since `Unix.realpath` is available in the OCaml versions that this project support, so once removed there should only be code with my copyright (Copyright 2024-2025 Kate Deplaix) + yours, which i'm happy to re-license to `BSD-1-Clause` or any similar permissive license for this project

shym:
> I thought there already is such a binary in the repository […]

@hannesm: IIUC, that binary is very limited (patching exactly one file, in particular), to run some tests only.

> i'm happy to re-license to BSD-1-Clause or any similar permissive license for this project

@kit-ty-kate: so ISC would suit you (they seem fairly close to my non-lawyer eyes)?

(I have a series of improvements in mind for `opatch`, which would remove in particular the `realpath` function)

hannesm:
> Does opatch really need to be in the same repository as `patch`? As far as i can see it only depends on its public interface

I honestly don't care much. If someone cares, please create a new repository and extract opatch, cut a release, and send a PR to the patch repository to remove opatch.

